PR #84, opened 2021-12-04
Create page with routing

dmtrhrytsak:
[Link](https://dmtrhrytsak.github.io/react_people-table-basics/)

### DIFF
--- a/src/lib/api/people.ts
+++ b/src/lib/api/people.ts
@@ -1,0 +1,21 @@
+import { Person } from '../../types/Person';
+
+const wait = (delay: number): Promise<typeof delay> => {
+  return new Promise((resolve) => {
+    setTimeout(() => resolve(delay), delay);
+  });
+};
+
+export const getPeople = async (): Promise<Person[]> => {
+  await wait(1000);
+
+  const response = await fetch(
+    'https://mate-academy.github.io/react_people-table/api/people.json',
+  );
+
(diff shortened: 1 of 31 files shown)


src/lib/api/people.ts
```diff
@@ -0,0 +1,21 @@
+import { Person } from '../../types/Person';
+
+const wait = (delay: number): Promise<typeof delay> => {
+  return new Promise((resolve) => {
+    setTimeout(() => resolve(delay), delay);
+  });
+};
+
+export const getPeople = async (): Promise<Person[]> => {
+  await wait(1000);
```
AlinaMatuschak, line 10: Why did you use timeout here? For spinner?

src/pages/HomePage.tsx
```diff
@@ -0,0 +1,11 @@
+import { Wrapper } from '../components/Wrapper';
+
+export const HomePage = () => {
+  return (
+    <section className="HomePage">
+      <Wrapper>
```
AlinaMatuschak, line 6: You can add a wrapper for the app instead add to every page

AlinaMatuschak (approved):
Good work 👌

src/lib/api/people.ts
dmtrhrytsak: Exactly! So that everyone can see it:)

PR #87, opened 2021-12-30
Develop

Val-space:
[People table api](https://Val-space.github.io/react_people-table-basics/)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@ node_modules
 .DS_Store
 yarn.lock
 package-lock.json
+package.json
 
 npm-debug.log*
 yarn-debug.log*
--- a/src/components/PeopleTable/PeopleTable.tsx
+++ b/src/components/PeopleTable/PeopleTable.tsx
@@ -1,0 +1,31 @@
+import 'bulma/css/bulma.min.css';
+import { PersonRow } from '../PersonRow/PersonRow';
+import './peopleTable.scss';
+
+type Props = {
+  people: Person[];
+};
+
+export const PeopleTable: React.FC<Props> = ({ people }) => {
+  return (
+    <>
+      <table className="PeopleTable table">
+        <thead>
+          <tr>
+            <th>Name</th>
+            <th>sex</th>
+            <th>born</th>
+            <th>died</th>
+            <th>mother</th>
+            <th>father</th>
+          </tr>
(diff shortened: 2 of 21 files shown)


.gitignore
```diff
@@ -5,6 +5,7 @@ node_modules
 .DS_Store
 yarn.lock
 package-lock.json
+package.json
```
dmtrhrytsak, line 8: Listing package.json and package-lock.json in here isn't the best idea😅 There're might be some situations when you do want and need to commit them

dmtrhrytsak (approved):
Bravo!

dmtrhrytsak:
When I open the page it's showing me this...
image

src/components/PeopleTable/PeopleTable.tsx
```diff
@@ -0,0 +1,31 @@
...
+import { PersonRow } from '../PersonRow/PersonRow';
+import './peopleTable.scss';
+
+type Props = {
+  people: Person[];
+};
+
+export const PeopleTable: React.FC<Props> = ({ people }) => {
+  return (
+    <>
```
AlinaMatuschak, line 11: This fragment is not necessary here

AlinaMatuschak (approved):

.gitignore
Val-space: nooo, Didn't you tell me not to add the package.json ..i'll check

Val-space:
> When I open the page it's showing me this... image

yeah i know, but it works on 3000 port, i don't know it's deploy error or something else

.gitignore
dmtrhrytsak: I guess it was me😅 But people get smarter and opinions change.. so let’s say I was just a little bit too dumb))

But, actually, I’ve always thought that we should commit these two files when we’ve installed new packages. This way whoever fork/clone this repo will install all the  necessary packages (responsible for that is `package.json`) with the same versions (  responsible for that is `package-lock.json`) (Commands to install `npm install/yarn install)`

dmtrhrytsak: And if I’m being honest, it was Illya who told us to not do that… So, I’m not the root of this problem😆

Val-space: Not only you, Dima, many people told me not to commit the file. I'm lost😁

PR #118, opened 2022-03-30
done

krasnobokydima:
[DEMO LINK](https://krasnobokydima.github.io/react_people-table-basics/)


src/api/users.ts
```diff
@@ -0,0 +1,9 @@
+export const users = async () => {
```
Elena-brv, line 1:
```suggestion
export const getUsers = async () => {
```

src/components/PeoplePage.tsx
```diff
@@ -0,0 +1,36 @@
+import React, { useEffect, useState } from 'react';
+import { users } from '../api/users';
+import { PeopleTable } from './PeopleTable';
+
+export const PeoplePage: React.FC = () => {
+  const [peopleFromServer, setPeopleFromServer] = useState<UserFromServerType[]>([]);
+
+  useEffect(() => {
+    users().then(usersFromServer => {
+      const newUsers = [...usersFromServer].map((userFromServer) => {
```
Elena-brv, line 10: map - no-mutation method no need to make a copy

src/components/PeopleTable.tsx
```diff
@@ -0,0 +1,23 @@
+import React from 'react';
+import { PersonRow } from './PersonRow';
+
+export const PeopleTable:
+React.FC<{ peopleFromServer: UserFromServerType[] }> = ({ peopleFromServer }) => (
```
Elena-brv, line 5: create a separate type Props

```suggestion
React.FC<Props> = ({ peopleFromServer }) => (
```

Elena-brv (changes requested):
- remove unnecessary files from PR
- here should be name of Repo also 
image

PR #165, opened 2022-05-27
Develop

Yurii837:


src/App.tsx
```diff
@@ -1,9 +1,31 @@
...
+      </Link>
+
+      <Link to="/peoples" className="nav-link">
+        Peoples
+      </Link>
+
+      <Switch>
+        <Route path="/" exact component={HomePage} />
+        <Route path="/peoples" component={PeoplePage} />
+        <Redirect path="/home" to="/" />
```
serhii-naumenko, line 23: You use Redirect, but you have never used "/home", so your Redirect is nonsensical. Though conditions of the task don't order it exactly.

src/Api.ts
```diff
@@ -0,0 +1,6 @@
+// eslint-disable-next-line max-len
+const apiLink = 'https://mate-academy.github.io/react_people-table/api/people.json';
+
+export const getPeople = async () => {
+  return (await fetch(apiLink)).json();
```
serhii-naumenko, line 5: I think it's better to divide awaiting from fetch and from method json.

src/react-app-env.d.ts
```diff
@@ -1,2 +1,9 @@
-// eslint-disable-next-line
-/// <reference types="react-scripts" />
+interface People {
```
serhii-naumenko, line 1: The name People doesn't correspond to sense. It's maybe better to name Person.

src/Components/PeoplePage/PeoplePage.tsx
```diff
@@ -0,0 +1,42 @@
...
+        : peoples
+          ? <PeopleTablePage peoples={peoples} />
+          : (
+            <Oval
+              height="200"
+              width="200"
+              color="grey"
+              ariaLabel="loading"
+            />
+          )}
```
MaksimusCRS, line 39: Quite difficult to understand such logic, create separate variables

src/Components/PeopleTablePage/PeopleTablePage.tsx
```diff
@@ -0,0 +1,33 @@
...
+  return (
+    <table className="PeopleTable">
+      <thead>
+        <tr className="table-header">
+          <th className="table-header__title">Name</th>
+          <th className="table-header__title">Sex</th>
+          <th className="table-header__title">Born</th>
+          <th className="table-header__title">Died</th>
+          <th className="table-header__title">Mother</th>
+          <th className="table-header__title">Father</th>
```
MaksimusCRS, line 19: use map

src/react-app-env.d.ts
MaksimusCRS: Yep, you have only 1 person here.

MaksimusCRS (approved):
GJ

PR #186, opened 2022-06-02
react-people-table-basics

xapg6acc:
[DEMO LINK](https://xapg6acc.github.io/react_people-table-basics/)


Bohdan-Maksymiuk (changes requested):
Screenshot 2022-06-03 at 11 46 38
why do you need 2 home pages here and Not found page here?
they should work but shouldn't be visible in tabs

src/components/Table/Table.tsx
```diff
@@ -0,0 +1,43 @@
...
+                <th
+                  className="mdc-data-table__header-cell"
+                  role="columnheader"
+                  scope="col"
+                >
+                  {item}
+                </th>
+              ))}
+
+            </tr>
```
marina-tilniak, line 30:
```suggestion
              ))}
            </tr>
```

marina-tilniak (approved):
Good job 🚀

PR #201, opened 2022-06-23
Developer

kitaminary:
[DEMO LINK](https://kitaminary.github.io/react_people-table-basics/)


src/App.tsx
```diff
@@ -1,9 +1,41 @@
+import React from 'react';
+import * as reactRouterDom from 'react-router-dom';
```
Anna-Tkachenko, line 2: You should export only NavLink and Route from reactRouterDom

src/App.tsx
```diff
@@ -1,9 +1,41 @@
...
+          Home
+        </reactRouterDom.NavLink>
+        {' '}
+        <reactRouterDom.NavLink
+          to="/people"
+          className="button m-5"
+        >
+          People page
+        </reactRouterDom.NavLink>
+      </nav>
```
Anna-Tkachenko, line 25: Will be better to create separate Header component with this navigation

Anna-Tkachenko (approved):

PR #213, opened 2022-06-26
first router practice

volodya-leskiv:
[DEMO LINK](https://volodya-leskiv.github.io/react_people-table-basics/)


src/components/PeopleTable/PeopleTable.tsx
```diff
@@ -0,0 +1,28 @@
...
+          <td>Sex</td>
+          <td>Born</td>
+          <td>Died</td>
+          <td>Father</td>
+          <td>Mather</td>
+        </tr>
+      </thead>
+      <tbody>
+        {people?.map(person => (
+          <PersonRow person={person} />
```
Anna-Tkachenko, line 23: Don't forget about key attribute

src/components/PersonRow/PersonRow.tsx
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
+
+type Props = {
+  person: Person | null
```
Anna-Tkachenko, line 4: Will be better to show this row only if person is not null

Anna-Tkachenko (approved):

PR #219, opened 2022-06-30
Develop

enrikadej:
[DEMO LINK](https://enrikadej.github.io/react_people-table-basics/)


src/index.tsx
```diff
@@ -1,8 +1,12 @@
 import ReactDOM from 'react-dom';
+import 'bulma/css/bulma.min.css';
+import { BrowserRouter as HashRouter } from 'react-router-dom';
```
Anna-Tkachenko, line 3: You need to use HashRouter, for now navigation doesn't work on demo link

Anna-Tkachenko (changes requested):

PR #229, opened 2022-07-10
Solution

amitpatiljc:
[DEMO LINK](https://amitpatiljc.github.io/react_people-table-basics/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,8 +1,68 @@
+import {
+  Navigate,
+  NavLink,
+  Route,
+  Routes,
+} from 'react-router-dom';
 import './App.scss';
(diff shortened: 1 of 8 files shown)


src/App.tsx
```diff
@@ -1,8 +1,68 @@
...
+              'is-tab',
+              { 'is-active': isActive },
+            )
+          )}
+          to="/people"
+        >
+          People Page
+        </NavLink>
+      </div>
+    </div>
```
59MrRobot, line 44: I'd suggest putting this in its own Header component

59MrRobot (changes requested):
Good job, just one suggestion

PR #231, opened 2022-07-13
task completed

andreisucman:
[DEMO LINK](https://andreisucman.github.io/react_people-table-basics/)


ihor-jpeg (changes requested):
The demo link does not work

PR #233, opened 2022-07-14
people table

Sardorbek1122:
[DEMO LINK](https://Sardorbek1122.github.io/react_people-table-basics/)

### DIFF
--- a/src/react-app-env.d.ts
+++ b/src/react-app-env.d.ts
@@ -1,2 +1,12 @@
 // eslint-disable-next-line
 /// <reference types="react-scripts" />
+
+interface People {
+  name: string;
+  sex: string;
+  born: number;
+  fatherName: string;
+  motherName: string;
+  slug: string;
+  died: number;
+}
--- a/README.md
+++ b/README.md
@@ -1,11 +1,11 @@
 # React - People table
 - Replace `<your_account>` with your Github username in the
-  [DEMO LINK](https://<your_account>.github.io/react_people-table-basics/)
+  [DEMO LINK](https://Sardorbek1122.github.io/react_people-table-basics/)
 - Follow the [React task guideline](https://github.com/mate-academy/react_task-guideline#react-tasks-guideline)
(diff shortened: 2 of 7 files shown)


src/components/PeopleTable.tsx
```diff
@@ -0,0 +1,28 @@
...
+        <th>Died</th>
+        <th>Mother</th>
+        <th>Father</th>
+      </tr>
+    </thead>
+    <tbody>
+      {people.map(person => (
+        <tr key={person.slug}>
+          <PersonRow person={person} />
+        </tr>
```
MabuzaM, line 24: I think it's ideal to keep this row inside `PersonRow` component and please rename `PeopleRow `to `PersonRow`. 

You can have your key passed to your component like this `<PersonRow person={person} key={person.slug}/>`

src/react-app-env.d.ts
```diff
@@ -1,2 +1,12 @@
 // eslint-disable-next-line
 /// <reference types="react-scripts" />
+
+interface People {
```
MabuzaM, line 4: Please name the type `Person`

MabuzaM (changes requested):
Good work. Have left some comments below. I also think you should keep your nav links at the same position when the user clicks on them. Currently they are centered when the home link is active and on the left when the people link is active.

ihor-jpeg (changes requested):
test

ihor-jpeg (changes requested):
test

ihorMate (changes requested):
test

ihorMate (changes requested):
test

ihor-jpeg (changes requested):
1

ihorMate (changes requested):
1

ihorMate (changes requested):
test

ihorMate (changes requested):
test

PR #238, opened 2022-07-21
React people table basics

vsupruniuk:
[DEMO LINK](https://vsupruniuk.github.io/react_people-table-basics/)

### DIFF
--- a/src/Components/PeoplePage/PeoplePage.tsx
+++ b/src/Components/PeoplePage/PeoplePage.tsx
@@ -1,0 +1,29 @@
+import React, { useEffect, useState } from 'react';
+import { Person } from '../../types/Person';
+import { getPeople } from '../../api/people';
+import { PeopleTable } from '../PeopleTable';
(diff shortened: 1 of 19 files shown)


src/Components/PeoplePage/PeoplePage.tsx
```diff
@@ -0,0 +1,29 @@
...
+      const response = await getPeople();
+
+      setPeople(response);
+    };
+
+    try {
+      fetchPeoples();
+    } catch (error) {
+      throw new Error(`${error}`);
+    }
```
jstmpelowycz, line 20: you could also use `catch` for promises, by doing so you avoid leaving promise returned by `fetchedPeoples()` just «in the air»
```
fetchPeople()
  .catch(error => console.log(error)); // or something like that
```
https://devtrium.com/posts/async-functions-useeffect

jstmpelowycz (approved):

PR #244, opened 2022-07-28
solution

liliia-livandovskaya:
[DEMO LINK](https://liliia-livandovskaya.github.io/react_people-table-basics/)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2121,9 +2121,9 @@
       }
     },
     "@mate-academy/scripts": {
-      "version": "1.2.0",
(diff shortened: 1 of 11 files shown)


AndriiNychaiuk (approved):
GJ! 🔥 
Almost everething is great, but I suggest a small fix. 
On change of the window size your [table content](https://prnt.sc/Sn1oyRXiqYs6) is transfering on the new line, it's not user friendly I guess.

Other thing is when I go on the link with table than for a short moment I see not centred head part of the table. I guess it's better to render whole table at the same time or add some loader.

Olexandr6 (approved):

PR #256, opened 2022-08-01
People table

AndrewKhorsun:
[DEMO LINK](https://andrewkhorsun.github.io/react_people-table-basics/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,9 +1,67 @@
 import './App.scss';
+import {
+  Routes, Route, NavLink, Navigate,
+} from 'react-router-dom';
+import { PeoplePage } from './components/PeoplePage';
+import { NotFound } from './components/NotFound';
 
-const App = () => (
-  <div className="App">
-    <h1>People table</h1>
-  </div>
-);
+const App = () => {
+  return (
+    <div className="App">
+      <div className="tabs is-centered">
+        <ul>
+          <li>
+            <NavLink
+              to="/"
+              style={({ isActive }) => {
+                return {
+                  color: isActive ? 'blue' : '',
+                };
+              }}
+            >
+              Home Page
+
+            </NavLink>
+          </li>
+          <li>
+            <NavLink
+              to="/PeoplePage"
+              style={({ isActive }) => {
+                return {
+                  color: isActive ? 'blue' : '',
+                };
+              }}
+            >
+              People Page
+
+            </NavLink>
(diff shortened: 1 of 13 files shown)


Angelion541 (approved):

src/App.tsx
```diff
@@ -1,9 +1,67 @@
...
+      <div className="tabs is-centered">
+        <ul>
+          <li>
+            <NavLink
+              to="/"
+              style={({ isActive }) => {
+                return {
+                  color: isActive ? 'blue' : '',
+                };
+              }}
```
jstmpelowycz, line 20: I believe using `classnames` would be a better solution instead of applying ternary with an empty string

src/App.tsx
```diff
@@ -1,9 +1,67 @@
...
+          element={<NotFound />}
+        />
+        <Route
+          path="/"
+          element={<p className="title">Home Page</p>}
+        />
+
+        <Route
+          path="/home"
+          element={<Navigate to="/" />}
```
jstmpelowycz, line 54: don't forget to add `replace`
here is why – https://reach.tech/router/api/navigate (at the bottom of the article)

jstmpelowycz (changes requested):

src/components/PeopleInfo.tsx
```diff
@@ -0,0 +1,17 @@
...
+};
+
+export const PeopleInfo: React.FC<Props> = ({ user }) => (
+  <>
+    <td>{user.name}</td>
+    <td>{user.sex}</td>
+    <td>{user.born}</td>
+    <td>{user.died}</td>
+    <td>{user.fatherName}</td>
+    <td>{user.motherName}</td>
```
mykola-koval, line 15: what if the user has no parent(s)?

src/components/PeoplePage.tsx
```diff
@@ -0,0 +1,61 @@
...
+import { PeopleInfo } from './PeopleInfo';
+
+const getPeople = ():Promise<User[]> => {
+  return (
+    fetch('https://mate-academy.github.io/react_people-table/api/people.json')
+      .then(res => res.json())
+      .catch(() => ({
+        Response: 'False',
+        Error: 'unexpected error',
+      }))
```
mykola-koval, line 13: this function probably should be in a separate file (`api`)

mykola-koval (approved):

PR #273, opened 2022-08-02
solution

vinnibea:
[DEMO LINK](https://<your_account>.github.io/react_people-table-basics/)

### DIFF
--- a/src/components/Navigation.tsx
+++ b/src/components/Navigation.tsx
@@ -1,0 +1,21 @@
+import React from 'react';
+import { NavLink } from 'react-router-dom';
+import 'bulma/css/bulma.min.css';
+import { Button } from 'react-bulma-components';
+
+export const Navigation: React.FC = () => {
+  return (
+    <nav className="App__nav">
+      <NavLink
+        to="/home"
+        className={({ isActive }) => (isActive ? 'is-active' : '')}
+      >
+        <Button>Home</Button>
+      </NavLink>
+
+      <NavLink to="/table">
+        <Button>People</Button>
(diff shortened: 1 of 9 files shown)


src/components/Navigation.tsx
```diff
@@ -0,0 +1,21 @@
...
+import { NavLink } from 'react-router-dom';
+import 'bulma/css/bulma.min.css';
+import { Button } from 'react-bulma-components';
+
+export const Navigation: React.FC = () => {
+  return (
+    <nav className="App__nav">
+      <NavLink
+        to="/home"
+        className={({ isActive }) => (isActive ? 'is-active' : '')}
```
jstmpelowycz, line 11: I believe using `classnames` would be a better solution instead of applying ternary with an empty string

src/components/Navigation.tsx
```diff
@@ -0,0 +1,21 @@
...
+      >
+        <Button>Home</Button>
+      </NavLink>
+
+      <NavLink to="/table">
+        <Button>People</Button>
+      </NavLink>
```
jstmpelowycz, line 18: why didn't you apply class names here as you did previously?

src/components/Notfoundpage.tsx
```diff
@@ -0,0 +1,7 @@
+export const NotFoundPage: React.FC = () => {
```
jstmpelowycz, line 1: fix you file name to `NotFoundPage`

jstmpelowycz (changes requested):

mykola-koval (approved):

PR #281, opened 2022-08-02
Develop

naziamoff:
[DEMO LINK](https://naziamoff.github.io/react_people-table-basics/)

### DIFF
--- a/src/components/PeoplePage.tsx
+++ b/src/components/PeoplePage.tsx
@@ -1,0 +1,65 @@
+import React, { useState, useEffect } from 'react';
+import classNames from 'classnames';
+import { Person } from '../types';
+import { getPeople } from '../api';
+
+export const PeoplePage: React.FC = () => {
+  const [people, setPeople] = useState<Person[]>([]);
+
+  useEffect(() => {
+    const fetchPeople = async () => (
+      setPeople(await getPeople())
+    );
+
+    try {
+      fetchPeople();
+    } catch {
+      throw new Error();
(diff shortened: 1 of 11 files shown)


src/components/PeoplePage.tsx
```diff
@@ -0,0 +1,65 @@
...
+import { Person } from '../types';
+import { getPeople } from '../api';
+
+export const PeoplePage: React.FC = () => {
+  const [people, setPeople] = useState<Person[]>([]);
+
+  useEffect(() => {
+    const fetchPeople = async () => (
+      setPeople(await getPeople())
+    );
```
Anna-Tkachenko, line 12: Will be better to create separate variable, store the result of getPeople function here, and then pass this variable to setPeople function.
Your code is short - but with separate variable it will be more understandable for other users

src/components/PeoplePage.tsx
```diff
@@ -0,0 +1,65 @@
...
+                {person.fatherName}
+              </td>
+              <td className={classNames({
+                'has-background-danger': !person.motherName,
+              })}
+              >
+                {person.motherName}
+              </td>
+            </tr>
+          ))}
```
Anna-Tkachenko, line 59: Good job, just create PeopleTable component, that will accept people array and render this table (according to the task)
image

Anna-Tkachenko (changes requested):

Bohdan-Maksymiuk (approved):

PR #298, opened 2022-08-23
whats wrong with loader ... or me

Havrysh:
https://havrysh.github.io/react_people-table-basics/


SanyaBratashchuk (changes requested):
DEMO Link is not working

PR #323, opened 2022-09-02
Solution

AndrewBessonoff:
[DEMO LINK](https://AndrewBessonoff.github.io/react_people-table-basics/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,170 +1,37 @@
-import { Loader } from './components/Loader';
+import {
+  Navigate,
+  Route,
+  Routes,
+} from 'react-router-dom';
+// import './App.css';
+import { NotFoundPage } from './components/NotFoundPage';
+import { Navbar } from './components/NavBar';
+import { HomePage } from './components/HomePage';
+import { People } from './components/PeoplePage';
 
 export const App = () => (
   <div data-cy="app">
-    <nav
-      data-cy="nav"
-      className="navbar is-fixed-top has-shadow"
-      role="navigation"
-      aria-label="main navigation"
-    >
-      <div className="container">
-        <div className="navbar-brand">
-          <a className="navbar-item" href="#/">
-            Home
-          </a>
-
-          <a
-            className="navbar-item has-background-grey-lighter"
-            href="#/people"
-          >
-            People
-          </a>
-        </div>
-      </div>
-    </nav>
-
+    <Navbar />
     <main className="section">
       <div className="container">
-        <h1 className="title">Home Page</h1>
-        <h1 className="title">People Page</h1>
-        <h1 className="title">Page not found</h1>
-
-        <div className="block">
-          <div className="box table-container">
-            <Loader />
-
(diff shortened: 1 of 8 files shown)


src/components/NavBar.tsx
```diff
@@ -0,0 +1,30 @@
...
+          <NavLink
+            to="/"
+            className={({ isActive }) => `navbar-item ${isActive ? 'has-background-grey-lighter' : ''}`}
+          >
+            Home
+          </NavLink>
+
+          <NavLink
+            to="/people"
+            className={({ isActive }) => `navbar-item ${isActive ? 'has-background-grey-lighter' : ''}`}
```
HennadiiKupr, line 22: And here

src/components/NavBar.tsx
```diff
@@ -0,0 +1,30 @@
...
+      data-cy="nav"
+      className="navbar is-fixed-top has-shadow"
+      role="navigation"
+      aria-label="main navigation"
+    >
+      <div className="container">
+        <div className="navbar-brand">
+          <NavLink
+            to="/"
+            className={({ isActive }) => `navbar-item ${isActive ? 'has-background-grey-lighter' : ''}`}
```
HennadiiKupr, line 15: You don't need here ternary operator. If isActive will be false, the class will be not added

HennadiiKupr (approved):

HennadiiKupr:
Nice job

src/App.tsx
```diff
@@ -1,170 +1,37 @@
...
+        <Routes>
+          <Route path="/" element={<HomePage />} />
+          <Route path="home" element={<Navigate to="/" replace />} />
+          <Route
+            path="people"
+            element={(
+              <>
+                <People />
+              </>
+            )}
```
ivan-kadykalo, line 26: Why do you need fragment here?
```suggestion
            element={(<People />)}
```

src/components/NavBar.tsx
ivan-kadykalo, line 15: Use classNames library in a whole project

src/components/NavBar.tsx
ivan-kadykalo, line 22: use classnames

And here is an example of how to do it without an empty class. But use classnames))
```suggestion
            className={({ isActive }) => `navbar-item ${isActive ?? 'has-background-grey-lighter'}`}
```

src/components/PeoplePage.tsx
```diff
@@ -0,0 +1,53 @@
...
+              Something went wrong
+            </p>
+          )}
+
+          {isEmpty && (
+            <p data-cy="noPeopleMessage">
+              There are no people on the server
+            </p>
+          )}
+          <PeopleTable people={people} />
```
ivan-kadykalo, line 48: If your people[] are empty, why do you need to render peopleTable?

ivan-kadykalo (changes requested):
GJ🚀 and almost done. Fix that comments.

ivan-kadykalo (approved):

PR #340, opened 2022-09-05
finished react-router task

Oleksandr-Pustovit:
[DEMO LINK](https://Oleksandr-Pustovit.github.io/react_people-table-basics/)


src/pages/PeoplePage.tsx
```diff
@@ -0,0 +1,55 @@
...
+  const [isLoading, setIsLoading] = useState(true);
+  const [people, setPeople] = useState<Person[] | null>(null);
+
+  useEffect(() => {
+    getPeople()
+      .then(res => {
+        setPeople(res);
+      })
+      .catch(() => {
+        //
```
illnino380, line 17: It's better to catch errors here. The empty array doesn't show problems with loading data

src/pages/PeoplePage.tsx
```diff
@@ -0,0 +1,55 @@
...
+                      There are no people on the server
+                    </p>
+                  ) : (
+                    <>
+                      <PeopleTable people={people} />
+                    </>
+                  )}
+                </>
+              )}
+            </>
```
illnino380, line 49: 1. Try to avoid nested ternary operators. It's hard to read all conditions
2. You should use the fragment only if the two elements are on the same level.
3. It's better to show a block with an error if you really caught one.

illnino380 (changes requested):
Good job! Just fix some issues.

src/pages/PeoplePage.tsx
Oleksandr-Pustovit: I trully agree with you... BUT if I throw an error here 1 test fails because it says I should not throw errors.
To be honest I was just tired to look for that test and try to understand where should  I operate that error to keep it green.

src/pages/PeoplePage.tsx
Oleksandr-Pustovit: 1. True. Should refactor that mess somehow.
2. My lint went crazy when I tried to leave that part without Fragment. It was sure I was missing some properties of actual jsx and I had to put some extra code there.
3. I think it depends on how the logic works. I intentionally gave people type Person | null to be able to control errors.

PR #354, opened 2022-10-06
Implement people basic table

Desertiks:
[DEMO LINK](https://Desertiks.github.io/react_people-table-basics/)


src/App.tsx
```diff
@@ -12,161 +18,26 @@ export const App = () => (
...
-                    <a href="#/people/philibert-haverbeke-1907">
-                      Philibert Haverbeke
-                    </a>
-                  </td>
-
-                  <td>m</td>
-                  <td>1907</td>
-                  <td>1997</td>
+        <Routes>
+          <Route path="/" element={<HomePage />} />
```
DimaKobzar7, line 30:
```suggestion
          <Route path="/" element={ <h1 className="title">Home Page</h1>} />
```
You don`t need to make component only for one tag/ And with the NotFoundPage the same situation

DimaKobzar7 (approved):
Good job!

src/App.tsx
dellachaise: @DimaKobzar7  I do not agree with you. Good practice here and for test tasks would be creating new components in such cases.

src/components/PeopleTable/PeopleTable.tsx
```diff
@@ -0,0 +1,112 @@
...
+
+                  return (
+                    <tr
+                      data-cy="person"
+                      key={slug}
+                      className={classNames(
+                        { 'has-background-warning': personSlug === slug },
+                      )}
+                    >
+                      <td>
```
dellachaise, line 61: Better create a separate component for nav item and use it here.

src/components/PeopleTable/PersonLink.tsx
```diff
@@ -0,0 +1,18 @@
...
+
+type Props = {
+  person: Person,
+};
+
+export const PersonLink: FC<Props> = ({ person }) => {
+  const { name, slug, sex } = person;
+
+  return (
+    <Link to={`../${slug}`} className={classNames({ 'has-text-danger': sex === 'f' })}>
```
dellachaise, line 14: Because `className` is long better rewrite it this way 👇  Would be easier to read.
```suggestion
    <Link 
      to={`../${slug}`} 
      className={classNames({ 'has-text-danger': sex === 'f' })}
    >
```

dellachaise (approved):
Good job 👍

PR #358, opened 2022-10-06
Solution

kir8mir:
- [DEMO LINK](https://kir8mir.github.io/react_people-table-basics/)


src/components/PeopleItem/PeopleItem.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+      <td>{sex}</td>
+      <td>{born}</td>
+      <td>{died}</td>
+      <td>
+        <NavLink
+          to={`/people/${findParrent(motherName || '-')}`}
+          style={findParrent(motherName || '-')
+            ? { pointerEvents: 'visible' }
+            : { pointerEvents: 'none', color: 'gray' }}
```
dellachaise, line 51: Why don't you create a class for this purpose and use it?

src/components/PeopleItem/PeopleItem.tsx
```diff
@@ -0,0 +1,77 @@
...
+        >
+          {motherName || '-'}
+        </NavLink>
+      </td>
+      <td>
+        <NavLink
+          to={`/people/${findParrent(fatherName || '-')}`}
+          style={findParrent(fatherName || '-')
+            ? { pointerEvents: 'visible' }
+            : { pointerEvents: 'none', color: 'gray' }}
```
dellachaise, line 67: Here the same.

src/components/PeopleList/PeopleList.tsx
```diff
@@ -0,0 +1,64 @@
...
+    setOnLoad(true);
+
+    getPeople()
+      .then(value => {
+        setPeople(value);
+        setPeopleLength(value.length);
+      })
+      .catch(() => setServerError(true))
+      .finally(() => setOnLoad(false));
+  }, [people.length]);
```
dellachaise, line 34: Let’s use `async await` all over the code. It gives you a simpler way to work with asynchronous promise-based code.

dellachaise (changes requested):
Good job 👍 

Just need to add some improvements according to the comments.

Something went wrong with your changes. Why do you have more than 2000 files changed? It shouldn't be this way.

Screenshot 2022-10-28 at 12 03 12

PR #359, opened 2022-10-07
implemented people table using React Router

yungdena:
[DEMO LINK](https://yungdena.github.io/react_people-table-basics/)

### DIFF
--- a/src/components/PeoplePage.tsx
+++ b/src/components/PeoplePage.tsx
@@ -1,0 +1,37 @@
+import { useEffect, useState } from 'react';
+import { getPeople } from '../api';
+import { Loader } from './Loader';
+import { Person } from '../types';
+import { PeopleTable } from './PeopleTable';
+
+export const PeoplePage = () => {
+  const [people, setPeople] = useState<Person[]>([]);
+  const [isLoading, setIsLoading] = useState(false);
+
+  const getPeopleFromServer = async () => {
+    try {
+      setIsLoading(true);
+      const response = await getPeople();
+
+      setPeople(response);
+    } finally {
+      setIsLoading(false);
+    }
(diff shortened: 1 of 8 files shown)


src/components/PeoplePage.tsx
```diff
@@ -0,0 +1,37 @@
...
+  }, []);
+
+  return (
+    <div className="block">
+      <h1 className="title">People Page</h1>
+
+      <div className="box table-container">
+        {isLoading && <Loader />}
+
+        {people && people.length > 0 && <PeopleTable people={people} />}
```
AlexanderKolomiiets, line 33: you have to handle cases when you have 0 people in your table and when you get an error

src/components/PersonLink.tsx
```diff
@@ -0,0 +1,70 @@
...
+          <Link
+            to={`/people/${father.slug}`}
+            className="has-text-danger"
+          >
+            {father.name}
+          </Link>
+        ) : (
+          fatherName || '-'
+        )}
+      </td>
```
AlexanderKolomiiets, line 67: your links do not work properly, they should reference to a parent in a people table

src/components/PersonLink.tsx
```diff
@@ -0,0 +1,70 @@
...
+          </Link>
+        ) : (
+          motherName || '-'
+        )}
+      </td>
+
+      <td>
+        {father ? (
+          <Link
+            to={`/people/${father.slug}`}
```
AlexanderKolomiiets, line 59: you cant get a slug from parent in this case. You should find an object in data with a same parent`s name.

AlexanderKolomiiets (changes requested):

PR #365, opened 2022-10-09
develop

YaroslavKolbaiev:
[DEMO LINK](https://YaroslavKolbaiev.github.io/react_people-table-basics/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,9 @@
-import { Loader } from './components/Loader';
-
+import {
+  NavLink, Route, Routes, Navigate,
+} from 'react-router-dom';
+import classNames from 'classnames';
 import './App.scss';
+import { PeoplePage } from './Pages.tsx';
 
 export const App = () => (
   <div data-cy="app">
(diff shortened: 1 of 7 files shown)


src/components/PeopleTable.tsx/PeopleTable.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+type Props = {
+  people: Person[]
+  selectedPerson: string
+};
+
+export const PeopleTable: React.FC<Props> = ({ people, selectedPerson }) => {
+  const existingPerson = (personsParent: string | null) => {
+    return people.find(person => person.name === personsParent);
+  };
```
stanislav-topikha, line 14:
```suggestion
  const getPersonByParent = (personsParent: string | null) => {
    return people.find(person => person.name === personsParent);
  };
```

src/Pages.tsx/PeoplePage.tsx
```diff
@@ -0,0 +1,57 @@
...
+        setLoading(false);
+        setShowPeople(true);
+      } catch (error) {
+        setLoading(false);
+        setDropError(true);
+      }
+    };
+
+    fetchPeople();
+  }, []);
```
Anastasiia-Svintsova, line 33: It's not a good practice to write an async function in useEffect(). Create a separate function for it and call it in useEffect. Also, you can use finally in try/catch construction to set Loading as false

src/components/PeopleTable.tsx/PeopleTable.tsx
```diff
@@ -0,0 +1,80 @@
...
+                <td>{person.born}</td>
+                <td>{person.died}</td>
+                <PersonLink
+                  parent={person.motherName}
+                  existingPerson={existingPerson}
+                />
+
+                <PersonLink
+                  parent={person.fatherName}
+                  existingPerson={existingPerson}
```
Anastasiia-Svintsova, line 70: it will be better to move it to the separate component

Anastasiia-Svintsova (changes requested):
Fix the comments please

src/components/PersonLink.tsx/PersonLink.tsx
```diff
@@ -0,0 +1,33 @@
...
+  return (
+    <td>
+      {!parent
+        && '-' }
+      {existingPerson(parent)
+        ? (
+          <Link
+            className={classNames(
+              { 'has-text-danger': existingPerson(parent)?.sex === 'f' },
+            )}
```
stanislav-topikha, line 24:
```suggestion
            className={classNames({
             'has-text-danger': existingPerson(parent)?.sex === 'f' 
            })}
```

src/components/PersonLink.tsx/PersonLink.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+export const PersonLink: React.FC<Props> = ({
+  parent,
+  existingPerson,
+}) => {
+  return (
+    <td>
+      {!parent
+        && '-' }
+      {existingPerson(parent)
```
stanislav-topikha, line 19: move it to the var, you used it 3 times

src/components/PersonLink.tsx/PersonLink.tsx
```diff
@@ -0,0 +1,33 @@
...
+};
+
+export const PersonLink: React.FC<Props> = ({
+  parent,
+  existingPerson,
+}) => {
+  return (
+    <td>
+      {!parent
+        && '-' }
```
stanislav-topikha, line 18:
```suggestion
}) => {
  if (!parent) {
    return (
     <td> - </td>
    );
  }
  
  return (
    <td>
```

would be a bit cleaner

stanislav-topikha (changes requested):

src/App.tsx
```diff
@@ -12,161 +15,39 @@ export const App = () => (
...
-                      Lieven de Causmaecker
-                    </a>
-                  </td>
-                </tr>
-              </tbody>
-            </table>
-          </div>
-        </div>
+        <Routes>
+          <Route path="*" element={<h1 className="title">Page not found</h1>} />
```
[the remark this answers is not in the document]
YaroslavKolbaiev: what is the sense of creating new component for a single row of code ?

src/Pages.tsx/PeoplePage.tsx
YaroslavKolbaiev: why is that ? in lessons coach saying that async fn takes more code but it's easier to understand those your are requesting

Anastasiia-Svintsova: I didn't say that the async function is bad. Just said that you need to move function declaration from useEffect() and just call it in useEffect().

src/components/PeopleTable.tsx/PeopleTable.tsx
```diff
@@ -0,0 +1,88 @@
...
+};
+
+export const PeopleTable: React.FC<Props> = ({ people, selectedPerson }) => {
+  const getPersonByParent = (personsParent: string | null) => {
+    return people.find(person => person.name === personsParent);
+  };
+
+  const activePerson = people.find(
+    person => person.slug === selectedPerson,
+  );
```
stanislav-topikha, line 18:
```suggestion
  const activePerson = people.find(person => 
    person.slug === selectedPerson,
  );
```

stanislav-topikha (approved):
GJ 👍🏻

PR #369, opened 2022-10-11
Develop

DimaKobzar7:
[DEMO LINK](https://DimaKobzar7.github.io/react_people-table-basics/)


src/App.tsx
```diff
@@ -1,173 +1,61 @@
...
+              path="home"
+              element={<Navigate to="/" replace />}
+            />
+            <Route path="/people">
+              <Route
+                // ставим индекс и он значит адрес что в обертке то есть /people
+                index
+                element={(
+                  <>
+                    {/* <h1 className="title">People Page</h1> */}
```
KorniychukLesya, line 36: You probably don't need these and other comments😏

src/App.tsx
```diff
@@ -1,173 +1,61 @@
...
+  return (
+    <div data-cy="app">
+      <Navbar />
+
+      <main className="section">
+        <div className="container">
+          <Routes>
+            <Route
+              path="/"
+              element={<h1 className="title">Home Page</h1>}
```
KorniychukLesya, line 24: Here, I would put it in a separate file, but the page has no context, so it can be left as is, I guess

KorniychukLesya (approved):
Good job!

PR #408, opened 2022-11-24
solution

mayuces:
[DEMO LINK](https://mayuces.github.io/react_people-table-basics/)


sapnachoudhary06 (changes requested):
Everything looks fine to me except the table head is visible even while the data is being loaded.

PR #419, opened 2022-12-12
Solution

vyurkov:
[DEMO LINK](https://vyurkov.github.io/react_people-table-basics/)


src/components/PeopleTable/PeopleTable.tsx
```diff
@@ -0,0 +1,63 @@
+import classNames from 'classnames';
+import React, { useCallback } from 'react';
+import { useParams } from 'react-router-dom';
+import { Person } from '../../types';
+import { PersonLink } from '../PersonLink';
+
+type Props = {
+  people: Person[] | undefined
```
vlad-demchuk, line 8: it shouldn't be `undefined`. It may be an empty array, but not `undefined`

src/components/PeopleTable/PeopleTable.tsx
```diff
@@ -0,0 +1,63 @@
...
+          >
+            <td>
+              <PersonLink person={person} />
+            </td>
+
+            <td>{person.sex}</td>
+            <td>{person.born}</td>
+            <td>{person.died}</td>
+            <td>{checkPerson(person.motherName)}</td>
+            <td>{checkPerson(person.fatherName)}</td>
```
vlad-demchuk, line 57: use destructuring for such cases

src/components/PersonLink/PersonLink.tsx
```diff
@@ -0,0 +1,21 @@
...
+  person: Person;
+};
+
+export const PersonLink: React.FC<Props> = ({ person }) => {
+  const { name, sex, slug } = person;
+
+  return (
+    <Link
+      className={classNames({ 'has-text-danger': sex === 'f' })}
+      to={`/people/${slug}`}
```
vlad-demchuk, line 16: you could use relative path instead of `people`

vlad-demchuk (changes requested):

TarasHoliuk (approved):

PR #427, opened 2022-12-19
solution

bradleymatjie:
[DEMO LINK](https://bradleymatjie.github.io/react_people-table-basics/)

### DIFF
--- a/src/components/PeopleTable.tsx
+++ b/src/components/PeopleTable.tsx
@@ -1,0 +1,174 @@
+import { FC } from 'react';
+import classNames from 'classnames';
+import { Person } from '../types';
+import { PersonLink } from './PersonLink';
+
+type Props = {
+  people: Person[];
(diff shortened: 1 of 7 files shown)


src/components/PeopleTable.tsx
```diff
@@ -0,0 +1,174 @@
...
+            <td>{findPerson(person.motherName)}</td>
+            <td>{findPerson(person.fatherName)}</td>
+          </tr>
+        ))}
+        <tr data-cy="person">
+          <td>
+            <a href="#/people/jan-van-brussel-1714">
+              Jan van Brussel
+            </a>
+          </td>
```
sapnachoudhary06, line 73: please, Remove these.

sapnachoudhary06 (changes requested):
Code looks good.
please fix hard coded values and the table head is visible even while the data is being loaded.

PR #444, opened 2022-12-21
Develop

Bibicutu:
[DEMO LINK](https://Bibicutu.github.io/react_people-table-basics/)


MeshackDev (approved):
Works well. One improvement would be to fix the loop when pressing the back button after visiting /home.

PR #457, opened 2022-12-28
solution

savchuksavchuk:
[DEMO LINK](https://savchuksavchuk.github.io/react_people-table-basics/)


src/components/Loader/PersonItem.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+      <td>{sex}</td>
+      <td>{born}</td>
+      <td>{died}</td>
+      <td>
+        <>
+          {(mother && typeof mother === 'object')
+            && <PersonLink person={mother} />}
+          {(mother && typeof mother === 'string') && `${mother}`}
+          {!mother && '-'}
```
polosanya, line 40: Can't you move this logic to findRelative function and just return a proper value from it to use it here?

src/components/Loader/PersonItem.tsx
```diff
@@ -0,0 +1,53 @@
...
+        </>
+      </td>
+      <td>
+        <>
+          {(father && typeof father === 'object')
+            && <PersonLink person={father} />}
+          {(father && typeof father === 'string') && `${father}`}
+          {!father && '-'}
```
polosanya, line 48: Same here

src/types/Error.ts
```diff
@@ -0,0 +1,5 @@
+export type Error = {
+  message: string,
+  type: string;
+  isDanger?: boolean,
+};
```
polosanya, line 5: https://unix.stackexchange.com/questions/18743/whats-the-point-in-adding-a-new-line-to-the-end-of-a-file
image

polosanya (changes requested):
Almost perfect!

Anastasiia-Svintsova (approved):

PR #464, opened 2023-01-05
solution

1stIndominus:
[DEMO LINK](https://1stIndominus.github.io/react_people-table-basics/)


src/components/PersonLink/PersonLink.tsx
```diff
@@ -0,0 +1,37 @@
...
+      {personIsParent ? (
+        <Link
+          className={classNames({
+            'has-text-danger': personIsParent?.sex === 'f',
+          })}
+          to={`/people/${personIsParent?.slug}`}
+        >
+          {parent}
+        </Link>
+      ) : parent}
```
loralevitska, line 34:
```suggestion
      ) :  (
        {parent}
      )}
```

loralevitska (changes requested):
Almost done)

src/components/PersonLink/PersonLink.tsx
1stIndominus: When I put this variable into curly braces than I've got such an error:

Type 'Element | { parent: string; }' is not assignable to type 'ReactNode'.
  Type '{ parent: string; }' is not assignable to type 'ReactNode'.
    Object literal may only specify known properties, and 'parent' does not exist in type 'ReactElement<any, string | JSXElementConstructor<any>> | ReactFragment | ReactPortal'.ts(2322)

src/components/PepleTable/PeopleTable.tsx
```diff
@@ -0,0 +1,86 @@
...
+import { PersonLink } from '../PersonLink/PersonLink';
+
+type Props = {
+  people: Person[],
+  selectedPerson: string,
+};
+
+export const PeopleTable: React.FC<Props> = ({ people, selectedPerson }) => {
+  const getPersonByParent = (parent: string | null) => {
+    return people.find(person => person.name === parent);
```
Anastasiia-Svintsova, line 13: It's not a good practice to leave the value undefined. Better use null here
```suggestion
    return people.find(person => person.name === parent) || null;
```

Anastasiia-Svintsova (approved):

PR #470, opened 2023-01-10
react_people-table-bas

DikanevYura:
https://DikanevYura.github.io/react_people-table-basics/


src/api.ts
```diff
@@ -1,6 +1,6 @@
 import { Person } from './types/Person';
 
-// eslint-disable-next-line max-len
```
TarasHoliuk, line 3: keep it to avoid lint error

src/components/PeopleTable/PeopleTable.tsx
```diff
@@ -0,0 +1,94 @@
...
+import { Person } from '../../types/Person';
+
+interface Props {
+  people: Person[];
+}
+
+export const PeopleTable: React.FC<Props> = ({ people }) => {
+  const { slug } = useParams();
+
+  const findParent = (parentName: string | null) => {
```
TarasHoliuk, line 13: It's not always `parent`, it could be the current person, so I'd name it `findPerson`
image

TarasHoliuk: Also, this function returns valid `JSX`, so you overwrite it as component `PersonLink`

TarasHoliuk (changes requested):
GJ :+1: I've added a few small comments to improve your code

nazarmatsevych (approved):

PR #473, opened 2023-01-11
add people table

iryna-borniak:
[DEMO LINK](https://iryna-borniak.github.io/react_people-table-basics/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,172 +1,25 @@
-import { Loader } from './components/Loader';
-
+import { Navigate, Route, Routes } from 'react-router-dom';
 import './App.scss';
+import { HomePage } from './components/HomePage';
+import { PeoplePage } from './components/PeoplePage';
+import { NotFoundPage } from './components/NotFoundPage';
+import { NavBar } from './components/NavBar';
 
 export const App = () => (
   <div data-cy="app">
-    <nav
-      data-cy="nav"
-      className="navbar is-fixed-top has-shadow"
-      role="navigation"
-      aria-label="main navigation"
-    >
-      <div className="container">
-        <div className="navbar-brand">
-          <a className="navbar-item" href="#/">
-            Home
-          </a>
-
-          <a
-            className="navbar-item has-background-grey-lighter"
-            href="#/people"
-          >
-            People
-          </a>
-        </div>
-      </div>
-    </nav>
+    <NavBar />
 
     <main className="section">
       <div className="container">
-        <h1 className="title">Home Page</h1>
-        <h1 className="title">People Page</h1>
-        <h1 className="title">Page not found</h1>
-
-        <div className="block">
-          <div className="box table-container">
-            <Loader />
-
-            <p data-cy="peopleLoadingError" className="has-text-danger">
-              Something went wrong
-            </p>
-
-            <p data-cy="noPeopleMessage">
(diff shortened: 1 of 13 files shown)


src/components/HomePage/HomePage.tsx
```diff
@@ -0,0 +1,3 @@
+export const HomePage: React.FC = () => (
+  <h1 className="title">Home Page</h1>
+);
```
OleksiiBlezniuk, line 3: Well, it's strange why linter didn't warn you about it, but we need to import React to use JSX syntax, also FC type can be imported from react lib too

OleksiiBlezniuk, line 3:
```suggestion
import React, { FC } from 'react';

export const HomePage: FC = () => (
  <h1 className="title">Home Page</h1>
);
```

src/App.tsx
```diff
@@ -1,172 +1,25 @@
...
-                  </td>
-                </tr>
-              </tbody>
-            </table>
-          </div>
-        </div>
+        <Routes>
+          <Route path="/" element={<HomePage />} />
+          <Route path="/home" element={<Navigate to="/" replace />} />
+          <Route path="people">
```
OleksiiBlezniuk, line 17: Would be nice to save routes to constant (object, or enum) to be sure that we using proper routes on other pages

src/components/PeoplePage/PeoplePage.tsx
```diff
@@ -0,0 +1,55 @@
...
+  const [people, setPeople] = useState<Person[]>([]);
+  const [isLoading, setIsLoading] = useState(false);
+  const [hasError, setHasError] = useState(false);
+
+  useEffect(() => {
+    const getPeopleFromServer = async () => {
+      try {
+        const peopleFromServer = await getPeople();
+
+        setIsLoading(true);
```
OleksiiBlezniuk, line 17: I guess this actions should be swapped in order, 'cause it's strange that we define that data is loading after this data loaded 🙂

src/components/PeoplePage/PeoplePage.tsx
```diff
@@ -0,0 +1,55 @@
...
+    getPeopleFromServer();
+  }, []);
+
+  return (
+    <>
+      <h1 className="title">People Page</h1>
+
+      <div className="block">
+        <div className="box table-container">
+          {!isLoading && <Loader />}
```
OleksiiBlezniuk, line 34: It's strange that we show loader when `isLoading` is `false`

OleksiiBlezniuk: Let's set `true` to `isLoading` before request, and `false` when this request returns value instead

src/components/PeoplePage/PeoplePage.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+          {hasError && (
+            <p data-cy="peopleLoadingError" className="has-text-danger">
+              Something went wrong
+            </p>
+          )}
+
+          {isLoading && !hasError && (
```
OleksiiBlezniuk, line 42: no need to define this condition here, you may just return loader or error text if this states has `true` value

OleksiiBlezniuk (changes requested):
Please make logic with loading more clear 🙂

PR #483, opened 2023-01-21
Develop

lefteris-L:
[DEMO LINK](https://lefteris-l.github.io/react_people-table-basics/)


src/App.tsx
```diff
@@ -1,172 +1,45 @@
...
-                </tr>
-              </tbody>
-            </table>
-          </div>
-        </div>
+        <Routes>
+          <Route
+            path="/home"
+            element={<Navigate to="/" replace />}
+          />
```
ysrckr, line 16: I think this should come after "/" route

src/components/PeoplePage/PeoplePage.tsx
```diff
@@ -0,0 +1,78 @@
...
+      console.error(error);
+
+      setHasError(true);
+    }
+  };
+
+  useEffect(() => {
+    getPeople();
+
+    return () => setPeople([]);
```
ysrckr, line 45: I did not understand why do you set people to empty array on unmount?

src/components/PeopleTable/PeopleTable.tsx
```diff
@@ -0,0 +1,76 @@
...
+import cn from 'classnames';
+import { Person } from '../../types';
+import { PersonLink } from '../PersonLink';
+
+type Props = {
+  people: Person[]
+};
+
+export const PeopleTable: FC<Props> = ({ people }) => {
+  const currentSlug = useMatch('/people/:slug')?.params.slug;
```
ysrckr, line 12: You can get the slug with useParams hook.
`const {slug} = useParams()`

ysrckr (approved):
Goog job 👏🏻

src/components/PeoplePage/PeoplePage.tsx
```diff
@@ -0,0 +1,78 @@
...
+import { PeopleTable } from '../PeopleTable';
+import { fetchPeople } from '../../api';
+import { Person } from '../../types';
+
+const getParents = (
+  { motherName, fatherName }: Person,
+  array: Person[],
+) => ({
+  mother: array.find(({ name }) => name === motherName),
+  father: array.find(({ name }) => name === fatherName),
```
naziamoff, line 14: what if parent is not found?

naziamoff (approved):

src/components/PeoplePage/PeoplePage.tsx
lefteris-L: So that it reloads on refresh

src/components/PeopleTable/PeopleTable.tsx
lefteris-L: What do I pass inside useParams?

src/components/PeoplePage/PeoplePage.tsx
ysrckr: Passing empty array to useEffect accomplish that anyway.  Normally you need to pass abort on unmount but for that you need to create an abortcontroller and pass it as a signal to fetch.

src/components/PeoplePage/PeoplePage.tsx
lefteris-L: null (I think) is passed by default and I make a ternary check in PeopleTable for it

src/components/PeopleTable/PeopleTable.tsx
ysrckr: You don't pass anything.  It return an object that has your parameters in that page so lets say you have `/users/:userid/:todoid` when you destructer it you get `const = { userid, todoid } = useParams()`
But be aware it returns string so if it is an id don;t forget to convert it to number.

src/components/PeoplePage/PeoplePage.tsx
lefteris-L: Yeah it works the same

PR #488, opened 2023-01-27
People table was implemented

vadiimvooo:


GabrielMaksimovich (changes requested):
GJ! Just fix one thing.

GabrielMaksimovich (approved):
Good< but something wrong with deploy

PR #492, opened 2023-01-28
https://frontend112.github.io/react_people-table-basics/

frontend112:


src/components/PeopleTable/PeopleTable.tsx
```diff
@@ -0,0 +1,206 @@
...
+export const PeopleTable: FC = () => {
+  const [people, setPeople] = useState<Person[]>([]);
+
+  const [error, setError] = useState('');
+
+  const [isLoading, setIsLoading] = useState(false);
+
+  const match = useMatch('/people/:selectedSlug');
+
+  const [selectedSlug, setSelectedSlug] = useState(match?.params.selectedSlug);
```
polosanya, line 26:
```suggestion
import { Person } from '../../types';
import { PersonLink } from '../PersonLink';
import { Loader } from '../Loader';
import { getPeople } from '../../api';

export const PeopleTable: FC = () => {
  const [people, setPeople] = useState<Person[]>([]);
  const [error, setError] = useState('');
  const [isLoading, setIsLoading] = useState(false);
  const match = useMatch('/people/:selectedSlug');
  const [selectedSlug, setSelectedSlug] = useState(match?.params.selectedSlug);
```
There is no need to add so many newlines. It makes your code harder to read

src/components/PeopleTable/PeopleTable.tsx
```diff
@@ -0,0 +1,206 @@
...
+    <>
+      {
+        isLoading
+          ? <Loader />
+          : (
+            <>
+              {
+                error
+                  ? (
+                    <p
```
polosanya, line 83:
```suggestion
          : ( { error
                  ? (<p
```
There is also no need to add fragments everywhere (<>).
And try to make your code cleaner. Format it correctly

polosanya (changes requested):
Let's make your code cleaner. Make sure that it's easy to read. You can press Shift + Alt + F to auto-format it correctly. 
Also, let's try to fix the problem with Eslint. Try deleting the `node_modules` folder and then typing `npm i` again. Ask for help in Slack chat if you have any difficulties)

frontend112:
@polosanya  I asked about that earlier in slack,
@qoa11a said that is fine

But today I tried today again to solve this,
install older version of typescript because lint shows info that newest is no supported, 
after installing older it shows error that needs newest version needed ;-(

polosanya (approved):

PR #499, opened 2023-01-31
Through bugs and lags

4ebupel:
[DEMO LINK](https://4ebupel.github.io/react_people-table-basics/)


Ihor-Karpyn (changes requested):
- I shouldn't see this message on loading
image

- after clicking on mother or father nothing happened 
- let's write more meaningful commit message
image

PR #508, opened 2023-02-01
solution

Sveta-Kryukova:
[DEMO LINK](https://Sveta-Kryukova.github.io/react_people-table-basics/)


src/index.tsx
```diff
@@ -1,14 +1,51 @@
...
+          <Route index element={<HomePage />} />
+
+          <Route path="/home" element={<Navigate to="/" replace />} />
+          <Route path="people">
+            <Route
+              index
+              element={(
+                <>
+                  <PeoplePage />
+                  <PeopleTable />
```
daryna-zinchenko, line 29: I am not sure if it has to be two separate components. I guess PeopleTable must be inside of PeoplePage. But maybe its not necessary.

daryna-zinchenko (approved):

daryna-zinchenko:
Looks great, i have only few optional suggestions.

src/index.tsx
solaryasha: Yes, PeopleTable should be inside `PeoplePage`

solaryasha (changes requested):

solaryasha (changes requested):
.

src/index.tsx
```diff
@@ -1,14 +1,39 @@
...
+    <HashRouter>
+      <Routes>
+        <Route path="/" element={<App />}>
+          <Route index element={<HomePage />} />
+
+          <Route path="/home" element={<Navigate to="/" replace />} />
+          <Route path="people">
+            <Route
+              index
+              element={(<PeoplePage />)}
```
solaryasha, line 25: You don't need the round brackets here
```suggestion
              element={<PeoplePage />}
```

solaryasha (approved):

PR #523, opened 2023-02-02
Implemented the App with 2 pages and ability to select a person in the table

viktoriia-kravchenko:
[DEMO LINK](https://viktoriia-kravchenko.github.io/react_people-table-basics/)


src/pages/PeoplePage.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+          {isPeopleLoadingError && (
+            <p data-cy="peopleLoadingError" className="has-text-danger">
+              Something went wrong
+            </p>
+          )}
+
+          {(isNoPeople && !isPeopleLoading) && (
+            <p data-cy="noPeopleMessage">
+              There are no people on the server
```
Yuriy-Khoptyanii, line 64: you have one problem with this - when I click PeoplePage, I see some blick.

this is a problem in 62 string 
if you want to fix this problem - you need one more state.
you should create isPeopleLoaded and change this state  - made 'true' when you try get people and use it instead isPeopleLoading in 62 string.

Yuriy-Khoptyanii (approved):
also good job!

solaryasha (approved):
Good job!

src/pages/PeoplePage.tsx
viktoriia-kravchenko: Thank you for such a detailed explanation! It worked for me. Haven't paid attention to this before you mentioned it.

PR #532, opened 2023-02-03
solution

Lun4trick:
[DEMO](https://lun4trick.github.io/react_people-table-basics/)
(is it a good practice to create the navbar like this?)

### DIFF
--- a/src/components/PeopleTable/PeopleTable.tsx
+++ b/src/components/PeopleTable/PeopleTable.tsx
@@ -1,0 +1,111 @@
+import { useEffect, useState } from 'react';
+import cn from 'classnames';
+import { useParams } from 'react-router-dom';
+import { getPeople } from '../../api';
+import { Person } from '../../types';
+import { Loader } from '../Loader';
+import { PersonLink } from '../PersonLink/PersonLink';
+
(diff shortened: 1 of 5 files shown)


src/components/PeopleTable/PeopleTable.tsx
```diff
@@ -0,0 +1,111 @@
...
+          </p>
+        )}
+
+        {isPeopleEmpty && (
+          <p data-cy="noPeopleMessage">
+            There are no people on the server
+          </p>
+        )}
+
+        {!isLoading && people?.length && (
```
UmizDemud, line 56: You could use the `isPeopleEmpty` here. One less data access // not consequential so you can just keep it this way

UmizDemud (approved):
Everything seems correct to me.

And yes, having that object for navbar would be handy. You can evolve it and call it site map in the future. It can also store test states, demo states

src/components/PeopleTable/PeopleTable.tsx
naziamoff: I'd create a variable for that

naziamoff (approved):

PR #543, opened 2023-02-09
Solution

anton-sosonskyi:
[DEMO LINK](https://anton-sosonskyi.github.io/react_people-table-basics/)


src/components/PeopleTable/PeopleTable.tsx
```diff
@@ -0,0 +1,75 @@
...
+                  <PersonLink person={person} />
+                </td>
+
+                <td>{sex}</td>
+                <td>{born}</td>
+                <td>{died}</td>
+                <td>
+                  {mother
+                    ? <PersonLink person={mother} />
+                    : motherName || '-'}
```
DarkMistyRoom, line 61: it isn't a good idea to use ternary with binary operators this way. think about refactoring. you can use helper variable

DarkMistyRoom (changes requested):
Good job, just one suggestion

ViktorHavryliuk (approved):

PR #552, opened 2023-02-16
Solution

CharAnagn:
[DEMO LINK](https://charanagn.github.io/react_people-table-basics/)


src/App.tsx
```diff
@@ -1,172 +1,37 @@
...
-              </tbody>
-            </table>
-          </div>
-        </div>
+        <Routes>
+          <Route
+            path="/"
+            element={<Homepage />}
+          />
+
```
Grayfox147, line 19:
```suggestion
 <Route
              path="/home"
              element={(
                <Navigate
                  to="/"
                  replace
                />
              )}
```
I believe you missed the Navigatate from the readme: Use [Navigate](https://reactrouter.com/docs/en/v6/components/navigate) component to redirect from `/home`

src/components/PeoplePage/PeoplePage.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+  useEffect(() => {
+    setIsLoading(true);
+    getPeople()
+      .then((result) => {
+        setIsLoading(false);
+        setPeople(result.map(person => ({
+          ...person,
+          mother: result
+            .find(mother => mother.name === person.motherName),
```
Grayfox147, line 24:
```suggestion
             const getParent = (name: string | null) => (
    people.find((parent) => parent.name === name) || null
  );
```
since we use the same code more than once I would recommend you to implement a helper function so you can use it in both cases, without typing twice.

src/components/PeoplePage/PeoplePage.tsx
```diff
@@ -0,0 +1,67 @@
...
+      })
+      .catch(() => {
+        setHasError(true);
+      })
+      .finally(() => {
+        setIsLoading(false);
+      });
+  }, []);
+
+  const noDataError = !isLoading && people.length === 0;
```
Grayfox147, line 37: Great job implementing a varialbe to save the conditions.

Grayfox147 (approved):
I believe everything should be working, although the Demo Link is not working. But overall everything should be working I just added some suggestions to improve the final code. Good Job!

src/components/Homepage/Homepage.tsx
```diff
@@ -0,0 +1,5 @@
+export const Homepage = () => {
+  return (
+    <h1 className="title">Home Page</h1>
+  );
+};
```
naziamoff, line 5: you can return right away here

src/components/Navigation/Navigation.tsx
```diff
@@ -0,0 +1,45 @@
...
+            className={({ isActive }) => (
+              cn(
+                'navbar-item',
+                {
+                  'has-background-grey-lighter': isActive,
+                },
+              )
+            )}
+            to="/"
+          >
```
naziamoff, line 24: lets save up some space
```suggestion
          <NavLink
            to="/"
            className={({ isActive }) => (
              cn('navbar-item', {
                'has-background-grey-lighter': isActive,
              }))}
          >
```

src/components/Navigation/Navigation.tsx
```diff
@@ -0,0 +1,45 @@
...
+            Home
+          </NavLink>
+
+          <NavLink
+            className={({ isActive }) => (
+              cn(
```
naziamoff, line 30: and here, it takes really a lot of space

src/components/PeoplePage/PeoplePage.tsx
naziamoff: yeah, agree with @Grayfox147 
But comparison with 0 looks kinda hardcoded, we usually do like that
```suggestion
  const noDataError = !isLoading && !people.length;
```

naziamoff (approved):
LGTM 👍

PR #557, opened 2023-02-18
Develop

mmahfuzi:
- [DEMO LINK](https://mmahfuzi.github.io/react_people-table-basics/)


src/components/PeopleTable.tsx
```diff
@@ -0,0 +1,101 @@
...
+    <div className="block">
+      {loading
+        ? (
+          <div className="box table-container"><Loader /></div>
+        )
+        : (
+          <div className="box table-container">
+            <p data-cy="peopleLoadingError" className="has-text-danger">
+              {errorMessage}
+            </p>
```
dahimohamed, line 32: the p tag should be rendered conditionally , only when there is an error ( inspect it and you will find that it is always there)

src/components/PeoplePage.tsx
```diff
@@ -0,0 +1,42 @@
...
+  const [errorMessage, setErrorMessage] = useState('');
+  const [loading, setLoading] = useState(true);
+
+  const match = useMatch('/people/:slug');
+  const selectedPersonSlug = match?.params.slug;
+
+  useEffect(() => {
+    setTimeout(() => {
+      setLoading(false);
+    }, 500);
```
dahimohamed, line 18: I don't understand why you are using here setTimeout(), instead you can do it like this (adding finally block) :
 const loadPeople = async () => {
    setIsLoading(true);

    try {
      const loadedPeople = await getPeople();

      setPeople(loadedPeople);
      setIsLoaded(true);
    } catch (error) {
      setHasError(true);
    } finally {
      setIsLoading(false);
    }
  };

  useEffect(() => {
    loadPeople();
  }, []);

src/components/PeopleTable.tsx
```diff
@@ -0,0 +1,101 @@
...
+                  const mother = parentName(person.motherName);
+                  const father = parentName(person.fatherName);
+
+                  return (
+                    <tr
+                      data-cy="person"
+                      key={person.slug}
+                      // onClick={() => {
+                      //   selectedPerson(person);
+                      // }}
```
dahimohamed, line 68: don't forget to delete those comments

dahimohamed (changes requested):
in general you did great, but you need to clean up !

src/components/PeopleTable.tsx
mmahfuzi: It is conditional with loading :)

src/components/Navigation.tsx
```diff
@@ -0,0 +1,42 @@
...
+  return (
+    <nav
+      data-cy="nav"
+      className="navbar is-fixed-top has-shadow"
+      role="navigation"
+      aria-label="main navigation"
+    >
+      <div className="container">
+        <div className="navbar-brand">
+
```
naziamoff, line 14: no empty spaces between parent and child elements

src/components/PeopleTable.tsx
```diff
@@ -0,0 +1,99 @@
...
+  errorMessage: string,
+  loading: boolean,
+  selectedPersonSlug: string | undefined,
+};
+
+export const PeopleTable: React.FC<Props> = (
+  {
+    people, errorMessage, loading, selectedPersonSlug,
+  },
+) => {
```
naziamoff, line 17:
```suggestion
export const PeopleTable: React.FC<Props> = ({
  people, 
  errorMessage, 
  loading, 
  selectedPersonSlug,
}) => {
```

src/components/PeopleTable.tsx
```diff
@@ -0,0 +1,99 @@
...
+    return people.find(p => p.name === name) || null;
+  }
+
+  return (
+    <div className="block">
+      {loading
+        ? (
+          <div className="box table-container"><Loader /></div>
+        )
+        : (
```
naziamoff, line 28:
```suggestion
      {loading
        ? (
          <div className="box table-container">
            <Loader />
          </div>
        )
        : (
```

src/components/PeopleTable.tsx
```diff
@@ -0,0 +1,99 @@
...
+                          <PersonLink person={father} />
+                        )
+                        : (<td>{person.fatherName || '-'}</td>)}
+                    </tr>
+                  );
+                })}
+              </tbody>
+            </table>
+          </div>
+        )}
```
naziamoff, line 96: it seems way too difficult to read, I got lost in those ternaries. 2 levels of nesting is already bad for ternaries, do you think this is possible to refactor?

naziamoff (changes requested):
looks good, but let's refactor those ternaries in the PeopleTable

src/components/PeopleTable.tsx
```diff
@@ -0,0 +1,105 @@
...
+
+                    <td>{person.sex}</td>
+                    <td>{person.born}</td>
+                    <td>{person.died}</td>
+
+                    {mother
+                      ? (
+                        <PersonLink person={mother} />
+                      )
+                      : (<td>{person.motherName || '-'}</td>)}
```
naziamoff, line 88: and the same with father
```suggestion
                    {mother
                       ? <PersonLink person={mother} />
                       : <td>{person.motherName || '-'}</td>}
```

src/components/PeopleTable.tsx
```diff
@@ -0,0 +1,105 @@
...
+                <th>Mother</th>
+                <th>Father</th>
+              </tr>
+            </thead>
+
+            <tbody>
+
+              {people.map(person => {
+                const mother = parentName(person.motherName);
+                const father = parentName(person.fatherName);
```
naziamoff, line 67:
```suggestion
                const mother = getParentName(person.motherName);
                const father = getParentName(person.fatherName);
```

naziamoff (approved):
great 👍

PR #588, opened 2023-02-25
Develop

Kravchenko71213:
https://Kravchenko71213.github.io/react_people-table-basics/


src/App.tsx
```diff
@@ -1,172 +1,24 @@
...
-                      Lieven de Causmaecker
-                    </a>
-                  </td>
-                </tr>
-              </tbody>
-            </table>
-          </div>
-        </div>
+        <Routes>
+          <Route path="*" element={<h1 className="title">Page not found</h1>} />
```
Anna-Tkachenko, line 14: let's move elements to separate components

src/components/PeopleTable/PeopleTable.tsx
```diff
@@ -0,0 +1,57 @@
...
+        </tr>
+      </thead>
+
+      <tbody>
+        {people.map(person => {
+          const mother = people
+            .find(personItem => personItem.name === person.motherName);
+          const father = people
+            .find(personItem => personItem.name === person.fatherName);
+
```
Anna-Tkachenko, line 43: will be better to add these objects to each person before calling `map` function
for now it'll be called for each rerendering, will be better to call it only one after receiving people from server

Anna-Tkachenko (approved):
Good job!

src/components/PeopleTable/PeopleTable.tsx
```diff
@@ -0,0 +1,49 @@
...
+import { Person } from '../../types';
+import { PersonItem } from '../PersonItem';
+
+type Props = {
+  people: Person[];
+  selectedSlug: string;
+};
+
+export const PeopleTable: FC<Props> = ({ people, selectedSlug }) => {
+  const isNoPeopleOnServer = people.length === 0;
```
Anna-Tkachenko, line 11: will be better not to show this component if there is no people

src/components/PersonItem/PersonItem.tsx
```diff
@@ -0,0 +1,56 @@
...
+export const PersonItem: FC<Props> = ({
+  people,
+  person,
+  selectedSlug,
+}) => {
+  const isSelectedPerson = person.slug === selectedSlug;
+  const personMotherName = person.motherName || '-';
+  const personFatherName = person.fatherName || '-';
+
+  const personMother = people
```
Anna-Tkachenko, line 21: will be better to prepare the array (add mother and father objects to each person) after retrieving people from server instead of on each component rendering

src/components/pages/PeoplePage/PeoplePage.tsx
```diff
@@ -0,0 +1,49 @@
...
+import { getPeople } from '../../../api';
+import { Person } from '../../../types';
+import { Loader } from '../../Loader';
+import { PeopleTable } from '../../PeopleTable';
+
+export const PeoplePage: React.FC = () => {
+  const [people, setPeople] = useState<Person[]>([]);
+  const [isError, setIsError] = useState(false);
+  const [isLoading, setIsLoading] = useState(false);
+  const { slug = '' } = useParams();
```
Anna-Tkachenko, line 12: do you need slug here?
if you used it just for passing as a props - delete it, and use from `useParams` in needed component

Anna-Tkachenko (changes requested):

src/components/PeopleTable/PeopleTable.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+    <tbody>
+      {people.map(person => (
+        <PersonItem
+          person={person}
+          key={person.slug}
+          mother={person.mother}
+          father={person.father}
+          motherName={person.motherName}
+          fatherName={person.fatherName}
```
Anna-Tkachenko, line 33: will be better to pass whole person object
probably further we need some other field from the person, and it will be easier to maintain it if we'll pass the whole object in `PersonItem`

Anna-Tkachenko (approved):

PR #598, opened 2023-02-27
Develop

Oleksandr-Maly:
[DEMO LINK](https://oleksandr-maly.github.io/react_people-table-basics/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,173 +1,22 @@
-import { Loader } from './components/Loader';
+import { Routes, Route, Navigate } from 'react-router-dom';
 
 import './App.scss';
+import { NavBar } from './components/NavBar';
+import { PeoplesPage } from './components/PeoplesPage';
 
 export const App = () => (
   <div data-cy="app">
-    <nav
-      data-cy="nav"
-      className="navbar is-fixed-top has-shadow"
-      role="navigation"
-      aria-label="main navigation"
-    >
-      <div className="container">
-        <div className="navbar-brand">
-          <a className="navbar-item" href="#/">
-            Home
-          </a>
-
-          <a
-            className="navbar-item has-background-grey-lighter"
-            href="#/people"
-          >
-            People
-          </a>
-        </div>
-      </div>
-    </nav>
+    <NavBar />
 
     <main className="section">
-      <div className="container">
-        <h1 className="title">Home Page</h1>
-        <h1 className="title">People Page</h1>
-        <h1 className="title">Page not found</h1>
-
-        <div className="block">
-          <div className="box table-container">
-            <Loader />
-
-            <p data-cy="peopleLoadingError" className="has-text-danger">
-              Something went wrong
-            </p>
-
-            <p data-cy="noPeopleMessage">
-              There are no people on the server
-            </p>
-
-            <table
-              data-cy="peopleTable"
-              className="table is-striped is-hoverable is-narrow is-fullwidth"
-            >
-              <thead>
-                <tr>
-                  <th>Name</th>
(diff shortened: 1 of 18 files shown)


src/components/PeoplesTable/PeopleTable.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+type Props = {
+  people: Person[]
+  selectedSlug:string,
+};
+
+export const PeopleTable: React.FC<Props> = ({ people, selectedSlug }) => {
+  const findPersonByName = (name:string | null) => {
+    return people.find(parent => parent.name === name);
+  };
```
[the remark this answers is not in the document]
Anna-Tkachenko: do you need this function here? if it need to be here only for passing as a props to the other component - it's not the best way of it's usage
let's move it somewhere in separate file
or it will be better to add `mother` and `father` properties for each person after retrieving them from the server

Anna-Tkachenko (changes requested):

src/App.tsx
```diff
@@ -1,173 +1,24 @@
...
-              </tbody>
-            </table>
-          </div>
-        </div>
-      </div>
+      <Routes>
+        <Route path="*" element={<NotFoundPage />} />
+        <Route path="/" element={<HomePage />} />
+        <Route path="home" element={<Navigate to="/" replace />} />
+        <Route path="/people" element={<PeoplesPage />}>
```
Anna-Tkachenko, line 18: let's add separate `route` with `index` property

src/components/PeoplesPage/PeoplesPage.tsx
```diff
@@ -0,0 +1,68 @@
+import { useEffect, useState } from 'react';
+import { useParams } from 'react-router-dom';
+import { getPeople } from '../../api';
+import { Person } from '../../types';
+import { findParentByName } from '../../utiles/findParentByName';
+import { Loader } from '../Loader';
+import { PeopleTable } from '../PeoplesTable';
+
+export const PeoplesPage = () => {
```
Anna-Tkachenko, line 9:
```suggestion
export const PeoplePage = () => {
```

src/components/PeoplesPage/PeoplesPage.tsx
```diff
@@ -0,0 +1,68 @@
...
+    fetchPeople();
+  }, []);
+  const isPeopleFromServer = !isLoaded && people.length === 0;
+
+  const prepearedPeople = people.map(person => {
+    const mother = findParentByName(people, person.motherName);
+    const father = findParentByName(people, person.fatherName);
+
+    return { ...person, mother, father };
+  });
```
Anna-Tkachenko, line 38: will be better to move it to separate util function
and transform people array only once after retrieving them from server instead of do this on each component rendering

src/components/PersonInfo/PersonInfo.tsx
```diff
@@ -0,0 +1,50 @@
...
+      data-cy="person"
+      className={classNames({ 'has-background-warning': isSelectedPerson })}
+    >
+      <PersonLink name={name} slug={slug} sex={sex} />
+
+      <td>{sex}</td>
+      <td>{born}</td>
+      <td>{died}</td>
+      {mother
+        ? <PersonLink name={mother.name} slug={mother.slug} sex={mother.sex} />
```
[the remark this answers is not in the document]
Anna-Tkachenko: you didn't fix the comment

Anna-Tkachenko (changes requested):

denys-cheporniuk (approved):

PR #600, opened 2023-02-27
Develop

OleksandrProkop:
[DEMO LINK](https://github.com/mate-academy/react_people-table-basics/pull/583)


src/App.tsx
```diff
@@ -1,173 +1,61 @@
...
-          </div>
+      </nav>
+
+      <main className="section">
+
+        <div className="container">
+          <Routes>
+            <Route
+              path="*"
+              element={<h1 className="title">Page not found</h1>}
```
Anna-Tkachenko, line 31: create separate component for each element

src/components/Navbar/Navbar.tsx
```diff
@@ -0,0 +1,28 @@
...
+    <div className="navbar-brand">
+      <NavLink
+        className={({ isActive }) => classNames(
+          'navbar-item',
+          { 'has-background-grey-lighter': isActive },
+        )}
+        to="/"
+      >
+        Home
+      </NavLink>
```
Anna-Tkachenko, line 15: let's move repeated code to the separate component

src/components/PersonList/PersonList.tsx
```diff
@@ -0,0 +1,73 @@
...
+          <th>Died</th>
+          <th>Mother</th>
+          <th>Father</th>
+        </tr>
+      </thead>
+
+      <tbody>
+        {personas.map(person => {
+          const mother = personas.find(mom => mom.name === person.motherName);
+          const father = personas.find(dad => dad.name === person.fatherName);
```
Anna-Tkachenko, line 36: what is `personas`?
will be better to add `mother` and `father` properties for each person after retrieving them from server
instead of searching objects on each callback call

src/components/PersonPage/PersonPage.tsx
```diff
@@ -0,0 +1,64 @@
...
+            : (
+              <>
+                {isError
+                  ? (
+                    <p data-cy="peopleLoadingError" className="has-text-danger">
+                      Something went wrong
+                    </p>
+                  ) : (
+                    <PersonList personas={personas} selectedPersonId={slug} />
+                  ) }
```
Anna-Tkachenko, line 51: don't use nested conditions

Anna-Tkachenko (changes requested):

src/components/PersonPage/PersonPage.tsx
OleksandrProkop: nested conditional isnt good, but other way worse.
if i want avoid nested, i must create variables for each message and event. And each variable have  individual combination conditions. I think undertanding this more hard then if you have some architecture with nested conditions

src/components/PersonList/PersonList.tsx
OleksandrProkop: sorry, naming is 80 level... my bad)

denys-cheporniuk (approved):

PR #610, opened 2023-02-28
Solution

max-rozzhalovets:


src/App.tsx
```diff
@@ -1,172 +1,28 @@
...
+        <Routes>
+          <Route path="/" element={<Home />} />
+          <Route path="/home" element={<Navigate to="/" replace />} />
+          <Route path="/people/">
+            <Route index element={<People />} />
+            <Route path=":userSlug" element={<People />} />
+          </Route>
+          <Route
+            path="/*"
+            element={<NotFound />}
```
Anna-Tkachenko, line 23:
```suggestion
          <Route
            path="/*"
            element={<NotFound />
          }
```

src/Pages/People/People.tsx
```diff
@@ -0,0 +1,38 @@
+import { useEffect, useState } from 'react';
+import { getPeople } from '../../api';
+import { PeopleTable } from '../../components/PeopleTable';
+import { Person } from '../../types';
+
+export const People = () => {
+  const [people, setPeople] = useState<Person[]>([]);
+  const [isSuccess, setIsSuccess] = useState<boolean>(true);
```
Anna-Tkachenko, line 8: don't add type for primitive values

src/Pages/People/People.tsx
```diff
@@ -0,0 +1,38 @@
...
+  const [isLoading, setIsLoading] = useState<boolean>(false);
+
+  const fetchPeople = async () => {
+    try {
+      setIsLoading(true);
+      const newPeople = await getPeople();
+
+      setPeople(newPeople);
+    } catch {
+      setIsSuccess(false);
```
Anna-Tkachenko, line 18: will be better to name the variable related to the error, instead of `success`
for now it's not clear why didn't you set `true` for this variable after successful response

src/components/NavBar/NavBar.tsx
```diff
@@ -0,0 +1,35 @@
...
+        <NavLink
+          className={({ isActive }) => classNames(
+            'navbar-item',
+            { 'has-background-grey-lighter': isActive },
+          )}
+          to="/"
+        >
+          Home
+        </NavLink>
+
```
Anna-Tkachenko, line 22: move repeated code in separate component

src/components/PeopleTable/PeopleTable.tsx
```diff
@@ -0,0 +1,134 @@
...
+                  fatherName,
+                } = person;
+
+                const mother = people.find((human) => (
+                  human.name === motherName
+                ));
+
+                const father = people.find((human) => (
+                  human.name === fatherName
+                ));
```
Anna-Tkachenko, line 77: let's add `mother` and `father` properties for each object after retrieving people from server instead of searching them on each callback call

Anna-Tkachenko (changes requested):

PR #613, opened 2023-03-02
s

grslckr:
[DEMO LINK](https://grslckr.github.io/react_people-table-basics/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,172 +1,35 @@
-import { Loader } from './components/Loader';
+import React from 'react';
 
 import './App.scss';
+import { Navigate, Route, Routes } from 'react-router-dom';
+import { NavBar } from './components/NavBar/NavBar';
+import { PeoplePage } from './components/PeoplePage/PeoplePage';
+import { Page } from './types/Page';
 
-export const App = () => (
+export const App: React.FC = () => (
   <div data-cy="app">
-    <nav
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -1,172 +1,35 @@
-import { Loader } from './components/Loader';
+import React from 'react';
```
naziamoff, line 2:
```suggestion
```

src/components/PeoplePage/PeoplePage.tsx
```diff
@@ -0,0 +1,69 @@
...
+        setPeopleFromServer(response.map(
+          person => {
+            const mother
+            = response.find(prsn => prsn.name === person.motherName);
+            const father
+            = response.find(prsn => prsn.name === person.fatherName);
+
+            return { ...person, mother, father };
+          },
+        ));
```
naziamoff, line 26: It looks really messy here. Let's do actions step by step and assign values we receive to a variable. I understand indeed the temptation to provide one-line solution, but making your code clear is usually valued much more that its shorteness

naziamoff: also, you don't want to have variables like `prsn` in your code

naziamoff (approved):
not bad. But I would suggest you to polish the method I commented, I see a space for improvements there

PR #624, opened 2023-03-20
Solution

Vasyl-Pavlenko:
[DEMO LINK](https://vasyl-pavlenko.github.io/react_people-table-basics/)


src/utils/findParents.ts
```diff
@@ -0,0 +1,5 @@
+import { Person } from '../types';
+
+export const findParents = (peopleData: Person[], name: string | null) => {
+  return peopleData.find(person => person.name === name);
```
mykhalenych, line 4:
```suggestion
export const findParents = (peopleData: Person[], name: string | null) => peopleData.find(person => person.name === name);
```

mykhalenych (changes requested):

PR #636, opened 2023-03-27
done

ghost:
[DEMO LINK](https://l4st1m0za.github.io/react_people-table-basics/)


src/components/People.tsx
```diff
@@ -0,0 +1,56 @@
...
+            <Loader />
+          )}
+
+          {error && (
+            <p data-cy="peopleLoadingError" className="has-text-danger">
+              Something went wrong
+            </p>
+          )}
+
+          {people.length === 0 && !error && !isLoading && (
```
OleziO, line 42: Replace with `!people.length` it will be the same as `people.length === 0`.

OleziO (changes requested):
You need to highlight clicked people row

l4st1m0za:
Rows does highlight, you just need to click right on the persons name

anstsot (approved):
Great job, approved!

PR #641, opened 2023-04-09
Added React People Table Basics

ctk12:
[DEMO LINK](https://ctk12.github.io/react_people-table-basics/)


src/App.tsx
```diff
@@ -1,172 +1,44 @@
...
-                  </td>
-                </tr>
-              </tbody>
-            </table>
-          </div>
-        </div>
+        <Routes>
+          <Route
+            path="/"
+            element={(<h1 className="title">Home Page</h1>)}
```
Anastasiia-Svintsova, line 17: It's better to create separate components for the home page and not found page

src/App.tsx
```diff
@@ -1,172 +1,44 @@
...
+
+          <Route
+            path="/people"
+            element={(<People />)}
+          />
+
+          <Route
+            path="/people/:slug"
+            element={(<People />)}
+          />
```
Anastasiia-Svintsova, line 35: Can't you rewrite it by using [this construction](https://reactrouter.com/en/main/route/route#index)?
image

src/components/People/People.tsx
```diff
@@ -0,0 +1,68 @@
...
+      }
+    } catch {
+      setIsError(true);
+    } finally {
+      setIsPeopleLoading(false);
+    }
+  };
+
+  useEffect(() => {
+    setIsPeopleLoading(true);
```
Anastasiia-Svintsova, line 34: Set it to true in the loadPeople function. Remove it from the useEffect

src/components/People/People.tsx
```diff
@@ -0,0 +1,68 @@
...
+import { PeopleTable } from '../PeopleTable';
+
+export const People: React.FC = () => {
+  const match = useMatch('people/:slug');
+  const selectedPersonSlug = match?.params.slug;
+
+  const [people, setPeople] = useState<Person[]>([]);
+  const [isPeopleLoading, setIsPeopleLoading] = useState(false);
+  const [isError, setIsError] = useState(false);
+  const [isNoPeople, setIsNoPeople] = useState(false);
```
Anastasiia-Svintsova, line 15: It's a redundant variable. Just check the people array length

src/components/PeopleTable/PeopleTable.tsx
```diff
@@ -0,0 +1,104 @@
...
+        {people.map(person => {
+          const {
+            name,
+            sex,
+            born,
+            died,
+            fatherName,
+            motherName,
+            slug,
+          } = person;
```
Anastasiia-Svintsova, line 40: Would be better to create a separate component for the person

src/components/PeopleTable/PeopleTable.tsx
```diff
@@ -0,0 +1,104 @@
...
+
+          const isMother = people.filter(personSearch => {
+            return personSearch.name === motherName;
+          })[0];
+
+          const isFather = people.filter(personSearch => {
+            return personSearch.name === fatherName;
+          })[0];
```
Anastasiia-Svintsova, line 48: Why do you need to use a filter here? Seems that the find method will be more appropriate. By the way, what will be if there is no mother or father? And `is` at the beginning means that it'll be a boolean type. But it's not, so rename it
```
const mother = people.find(personSearch => personSearch.name === motherName) || null;

Anastasiia-Svintsova (changes requested):
- The demo link is not working, make sure you deploy your changes 
image

Anastasiia-Svintsova (changes requested):
Pay attention to my comments and run deploy command before requesting a review

src/App.tsx
witflash: @ctk12 
Hi!
Did you have any difficulties with this approach?
You feel free to ask the question in Slack if you have.

src/components/People/People.tsx
witflash: Take attention to this comment and try to fix it.

witflash (changes requested):
Hi!
Pay attention to unresolved comments by prev mentors. And try to fix all the problems.

anstsot (approved):
Great job, approved!

PR #651, opened 2023-04-13
finished task

valeriiayarzhemska:
[DEMO LINK](https://valeriiayarzhemska.github.io/react_people-table-basics/)

### DIFF
--- a/src/components/NavigationBar/NavigationBar.tsx
+++ b/src/components/NavigationBar/NavigationBar.tsx
@@ -1,27 +1,25 @@
 import { FC } from 'react';
 import { NavigationLink } from '../NavigationLink';
 
-export const NavigationBar: FC = () => {
-  return (
-    <nav
-      data-cy="nav"
-      className="navbar is-fixed-top has-shadow"
-      role="navigation"
-      aria-label="main navigation"
-    >
-      <div className="container">
-        <div className="navbar-brand">
-          <NavigationLink
-            link="/"
-            title="Home"
-          />
+export const NavigationBar: FC = () => (
+  <nav
+    data-cy="nav"
+    className="navbar is-fixed-top has-shadow"
+    role="navigation"
+    aria-label="main navigation"
+  >
+    <div className="container">
+      <div className="navbar-brand">
+        <NavigationLink
+          link="/"
+          title="Home"
+        />
 
-          <NavigationLink
-            link="/people"
-            title="People"
-          />
-        </div>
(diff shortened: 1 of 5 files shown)


arkushenko-anna (approved):
GJ!

src/components/NavigationBar/NavigationBar.tsx
```diff
@@ -0,0 +1,27 @@
+import { FC } from 'react';
+import { NavigationLink } from '../NavigationLink';
+
+export const NavigationBar: FC = () => {
+  return (
```
OleksiiBlezniuk, line 5:
```suggestion
export const NavigationBar: FC = () => (
```

src/components/PeopleTable/PeopleTable.tsx
```diff
@@ -0,0 +1,86 @@
...
+            sex,
+            born,
+            died,
+            motherName,
+            fatherName,
+            slug,
+          } = person;
+
+          const motherOfPerson = findParent(people, motherName);
+          const fatherOfPerson = findParent(people, fatherName);
```
OleksiiBlezniuk, line 47: Let's prepare array to working with it and save needed fields there instead of calling `findParent` on every render

OleksiiBlezniuk (changes requested):
Let's show only one error here (if there was an error - show message about error, if array empty - about empty array, but not both at the same time)
Also let's render table only if there is data (no errors and array is not empty)
Screenshot 2023-04-19 at 18 13 45

src/helpers/helpers.ts
```diff
@@ -0,0 +1,11 @@
...
+
+export const findParent = (
+  people: Person[],
+) => (
+  people.map(person => ({
+    ...person,
+    mother: people.find((mother) => mother.name === person.motherName) || null,
+    father: people.find((father) => father.name === person.fatherName) || null,
+  }))
+);
```
OleksiiBlezniuk, line 11:
```suggestion
export const getPeopleWithParents = (
  people: Person[],
) => (
  people.map(person => ({
    ...person,
    mother: people.find((mother) => mother.name === person.motherName) || null,
    father: people.find((father) => father.name === person.fatherName) || null,
  }))
);
```

OleksiiBlezniuk: From findParent I expect that this function will find parent object only

OleksiiBlezniuk (approved):

PR #655, opened 2023-04-13
React people table

Anton-Tretiak:
[DEMO LINK](https://anton-tretiak.github.io/react_people-table-basics/)
image

### DIFF
--- a/src/components/PeoplePage/PeoplePage.tsx
+++ b/src/components/PeoplePage/PeoplePage.tsx
@@ -1,0 +1,63 @@
+import { useState, useEffect } from 'react';
+import { useParams } from 'react-router-dom';
+
+import { Loader } from '../Loader';
+import { PeopleTable } from '../PeopleTable/PeopleTable';
+
+import { getPeople } from '../../api';
+import { Person } from '../../types';
+
+export const PeoplePage = () => {
+  const [people, setPeople] = useState<Person[]>([]);
+  const [isLoading, setIsLoading] = useState(false);
+  const [isError, setIsError] = useState(false);
+  const { slug = '' } = useParams();
+
+  const loadPeople = async () => {
+    try {
+      const peopleFromServer = await getPeople();
+
+      setPeople(peopleFromServer);
+    } catch {
+      setIsError(true);
+    } finally {
+      setIsLoading(false);
+    }
+  };
+
+  useEffect(() => {
+    setIsLoading(true);
+
+    loadPeople();
+  }, []);
+
+  return (
+    <>
+      <h1 className="title">People Page</h1>
+
+      <div className="block">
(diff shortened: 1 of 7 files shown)


gretoq (approved):
GJ

src/components/PeoplePage/PeoplePage.tsx
```diff
@@ -0,0 +1,63 @@
...
+import { PeopleTable } from '../PeopleTable/PeopleTable';
+
+import { getPeople } from '../../api';
+import { Person } from '../../types';
+
+export const PeoplePage = () => {
+  const [people, setPeople] = useState<Person[]>([]);
+  const [isLoading, setIsLoading] = useState(false);
+  const [isError, setIsError] = useState(false);
+  const { slug = '' } = useParams();
```
OleksiiBlezniuk, line 14: better to add type for params and pass it here as generic

src/components/PeopleTable/PeopleTable.tsx
```diff
@@ -0,0 +1,84 @@
...
+            sex,
+            born,
+            died,
+            fatherName,
+            motherName,
+            slug,
+          } = person;
+
+          const mother = findParent(people, motherName);
+          const father = findParent(people, fatherName);
```
OleksiiBlezniuk, line 47: Better to prepare array with data for working with it and save this objects there instead of doing it on every render

src/components/PeopleTable/PeopleTable.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+          return (
+            <tr
+              data-cy="person"
+              className={classNames({
+                'has-background-warning': slug === id,
```
OleksiiBlezniuk, line 53: how can slug be equal to id, it's absolutely different values, use more descriptive names

src/components/PeoplePage/PeoplePage.tsx
```diff
@@ -0,0 +1,63 @@
...
+                  </p>
+                )}
+
+                {people.length === 0 && (
+                  <p data-cy="noPeopleMessage">
+                    There are no people on the server
+                  </p>
+                )}
+
+                <PeopleTable people={people} id={slug} />
```
OleksiiBlezniuk, line 56: Slug - is not id

OleksiiBlezniuk: And why do you need to pass it in props? just use `useParams` inside `PeopleTable`

OleksiiBlezniuk (changes requested):
Let's render table only when there is no errors
Also let's show only one error message at one time, if there was error while loading - show error about it, not about array is empty
Screenshot 2023-04-19 at 12 21 37

PR #661, opened 2023-04-14
Solution

arkushenko-anna:
[DEMO LINK](https://arkushenko-anna.github.io/react_people-table-basics/)


src/components/PeopplePage/index.ts
```diff
@@ -0,0 +1 @@
+export * from './PeoplePage';
```
Oleksandr-kopaevich, line 1: There is a typo in the folder name 'PeopplePage'. It should be 'PeoplePage' to maintain consistency with the file name 'PeoplePage'. 📝

src/App.tsx
```diff
@@ -12,161 +16,36 @@ export const App = () => (
...
-                      Lieven de Causmaecker
-                    </a>
-                  </td>
-                </tr>
-              </tbody>
-            </table>
-          </div>
-        </div>
+        <Routes>
+          <Route path="/" element={<HomePage />} />
```
Oleksandr-kopaevich, line 41: It's nice practice to have all your routes in some variable, for example, enum
```
export enum PATH {
  Main = '/',
  Home = '/home',
  People = '/people',
  Error = '/*',
  USER_ID = ':slug',
} 
```

src/components/PeopplePage/PeoplePage.tsx
```diff
@@ -0,0 +1,138 @@
...
+import { getPeople } from '../../api';
+import { Person } from '../../types';
+import { Loader } from '../Loader';
+
+export const PeoplePage = () => {
+  const [people, setPeople] = useState<Person[]>([]);
+  const { slug: selectedPersonSlug } = useParams();
+
+  useEffect(() => {
+    const getPeopleFromServer = async () => {
```
Oleksandr-kopaevich, line 13: I would move the function out `getPeopleFromServer` from useEffect and PeoplePage component and make it to return `peopleWithParents`. Then your useEffect would be small and easy to read like this:
```
  useEffect(() => {
    getPeopleFromServer().then(setPeople);
  }
```
But your variant is also valid. The only mistake it's that you didn't put `setPeople` as a dependency in the dependency array. React must have shown you a warning about that.

Oleksandr-kopaevich: Also, you need to have error handling for all requests. Open your demo on the home page, hard refresh, open dev tools, set the network throttling as Offline, and click on the People page. You'll get an error and your users will see endless loading, which is super confusing for them. It's good to show some message explaining what exactly happened.

src/components/PeopplePage/PeoplePage.tsx
```diff
@@ -0,0 +1,138 @@
...
+            </tbody>
+          </table>
+
+        )
+        : (
+          <div className="block">
+            <div className="box table-container">
+              <Loader />
+
+              {/* <p data-cy="peopleLoadingError" className="has-text-danger">
```
Oleksandr-kopaevich, line 125: This looks like unfinished logic of error handling. Maybe you forgot about it? Please make it to work or remove commented the code. A repository shouldn't have commented code.

src/components/PeopplePage/PeoplePage.tsx
Oleksandr-kopaevich: Try to add one more state variable using a state that will show if a request is loading or no. Every request usually goes with 3 states:
- loader state
- data state
- error state

src/components/PeopplePage/PeoplePage.tsx
```diff
@@ -0,0 +1,138 @@
...
+    };
+
+    getPeopleFromServer();
+  }, []);
+
+  return (
+    <>
+      <h1 className="title">People Page</h1>
+
+      {people.length
```
Oleksandr-kopaevich, line 47: According to the task checklist table below should be in a separate component.
`8. Pass people to PeopleTable component and render as a table (see the given markup);`

src/components/PeopplePage/PeoplePage.tsx
```diff
@@ -0,0 +1,138 @@
...
+                    </td>
+
+                    <td>{person.sex}</td>
+                    <td>{person.born}</td>
+                    <td>{person.died}</td>
+                    <td>
+                      {person.mother
+                        ? (
+                          <Link
+                            to={`../${person.mother?.slug}`}
```
Oleksandr-kopaevich, line 92: `person.mother` is always defined in this case because you have condition higher in the code, you don't need `?`

src/components/PeopplePage/PeoplePage.tsx
```diff
@@ -0,0 +1,138 @@
...
+                            className={classNames(
+                              { 'has-text-danger': person.mother.sex === 'f' },
+                            )}
+                          >
+                            {person.mother?.name}
```
Oleksandr-kopaevich, line 97: The same as above, don't use operators or functions when they don't change anything or don't make sense, it will confuse others.

src/components/PeopplePage/PeoplePage.tsx
```diff
@@ -0,0 +1,138 @@
...
+                      </Link>
+                    </td>
+
+                    <td>{person.sex}</td>
+                    <td>{person.born}</td>
+                    <td>{person.died}</td>
+                    <td>
+                      {person.mother
+                        ? (
+                          <Link
```
Oleksandr-kopaevich, line 91: According to point 9.1 in the task checklist it must be a separate component
`create the <PersonLink person={person} /> and use it for all existing people;`

Oleksandr-kopaevich (changes requested):
Hey @arkushenko-anna demo looks great and works fine for the positive flow. You just need to handle negative flow (it's when some errors can happen) and refactor your code according to the task checklist (create 2 components PeopleTable and PersonLink and move existing code there). Don't give up, just push a little bit more and the task is 100% done.

Oleksandr-kopaevich:
And please, check why some checks failed.
[Test / run_tests (14.x) (pull_request) Failing after 4m](https://github.com/mate-academy/react_people-table-basics/actions/runs/4697988563/jobs/8329714726?pr=661)

PR #664, opened 2023-04-15
Implement table with people info MVP

gretoq:
[Demo link](https://gretoq.github.io/react_people-table-basics/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,173 +1,30 @@
-import { Loader } from './components/Loader';
+import { Navigate, Route, Routes } from 'react-router-dom';
 
 import './App.scss';
 
-export const App = () => (
-  <div data-cy="app">
-    <nav
-      data-cy="nav"
-      className="navbar is-fixed-top has-shadow"
-      role="navigation"
-      aria-label="main navigation"
-    >
-      <div className="container">
-        <div className="navbar-brand">
-          <a className="navbar-item" href="#/">
-            Home
-          </a>
+import { PeoplePage } from './pages/PeoplePage';
+import { NotFoundPage } from './pages/NotFoundPage';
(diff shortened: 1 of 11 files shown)


src/App.tsx
```diff
@@ -1,173 +1,28 @@
-import { Loader } from './components/Loader';
```
OleksiiBlezniuk, line 1: The Loader component import has been removed but it might be useful to show a loading indicator when fetching data from the API. Consider adding the Loader back and using it in the PeoplePage component. 😊

src/components/PeopleTable.tsx
```diff
@@ -0,0 +1,41 @@
+import { useParams } from 'react-router-dom';
+import { Person } from '../types';
+import { PersonRowInfo } from './PersonRowInfo';
+
+interface Props {
+  people: Person[];
+  // selectedSlug: string,
```
OleksiiBlezniuk, line 7: Remove commented code

src/components/PeopleTable.tsx
```diff
@@ -0,0 +1,41 @@
...
+}
+
+export const PeopleTable: React.FC<Props> = ({ people }) => {
+  // const { people, selectedSlug } = props;
```
OleksiiBlezniuk, line 11: Comment is not needed here as it's clear enough what the code is doing. Please remove it. 🧹

src/components/PeopleTable.tsx
```diff
@@ -0,0 +1,41 @@
...
+  const { slug: selectedSlug = '' } = useParams();
```
OleksiiBlezniuk, line 12: Better to add type for params and pass it as generic here

OleksiiBlezniuk (approved):

PR #665, opened 2023-04-16
create react people table

BozhenaBodnar:
[DEMO LINK](https://BozhenaBodnar.github.io/react_people-table-basics/)

### DIFF
--- a/src/components/NavBar/NavBar.tsx
+++ b/src/components/NavBar/NavBar.tsx
@@ -1,0 +1,17 @@
+import { PageNavLink } from '../../PageNavLink';
+
+export const NavBar = () => (
+  <nav
+    data-cy="nav"
+    className="navbar is-fixed-top has-shadow"
+    role="navigation"
+    aria-label="main navigation"
+  >
+    <div className="container">
+      <div className="navbar-brand">
+        <PageNavLink to="/" text="Home" />
+        <PageNavLink to="people" text="People" />
+      </div>
+    </div>
+  </nav>
+);
--- a/src/components/PeopleTable/PeopleTable.tsx
+++ b/src/components/PeopleTable/PeopleTable.tsx
@@ -1,0 +1,89 @@
+import { FC } from 'react';
+import classNames from 'classnames';
+import { Person } from '../../types';
+import { PersonLink } from '../PersonLink';
+
+type Props = {
+  people: Person[];
+  selectedPerson: string;
+};
+
+export const PeopleTable: FC<Props> = ({ people, selectedPerson }) => {
+  const isSelected = (person: Person) => person.slug === selectedPerson;
+
+  return (
+    <table
+      data-cy="peopleTable"
+      className="table is-striped is-hoverable is-narrow is-fullwidth"
+    >
+      <thead>
+        <tr>
+          <th>Name</th>
+          <th>Sex</th>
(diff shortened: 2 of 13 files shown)


src/components/NavBar/NavBar.tsx
```diff
@@ -0,0 +1,17 @@
...
+  <nav
+    data-cy="nav"
+    className="navbar is-fixed-top has-shadow"
+    role="navigation"
+    aria-label="main navigation"
+  >
+    <div className="container">
+      <div className="navbar-brand">
+        <PageNavLink to="/" text="Home" />
+        <PageNavLink to="people" text="People" />
```
OleksiiBlezniuk, line 13: Change the 'to' prop value to '/people' to ensure correct routing. 🚀

src/pages/PeoplePage.tsx
```diff
@@ -0,0 +1,72 @@
...
+import { Person } from '../types';
+import { PeopleTable } from '../components/PeopleTable';
+import { getPeople } from '../api';
+
+export const PeoplePage: FC = () => {
+  const [people, setPeople] = useState<Person[]>([]);
+  const [isLoading, setIsLoading] = useState(false);
+  const [hasError, setHasError] = useState(false);
+
+  const { personSlug = '' } = useParams();
```
OleksiiBlezniuk, line 19: The type of 'personSlug' in useParams should be explicitly specified. Consider using 'useParams<{ personSlug?: string }>()'

src/components/PeopleTable/PeopleTable.tsx
```diff
@@ -0,0 +1,87 @@
...
+import { Person } from '../../types';
+import { PersonLink } from '../PersonLink/PersonLink';
+
+type Props = {
+  people: Person[];
+  selectedPerson: string;
+};
+
+export const PeopleTable: FC<Props> = ({ people, selectedPerson }) => {
+  const isSelected = (person: Person) => person.slug === selectedPerson;
```
OleksiiBlezniuk, line 12: Why can't we save this to constant inside map method?

src/components/PeopleTable/PeopleTable.tsx
```diff
@@ -0,0 +1,87 @@
...
+                  ? (
+                    <PersonLink
+                      person={
+                        people.find(
+                          woman => woman.name === motherName,
+                        ) || person
+                      }
+                    />
+                  )
+                  : motherName || '-'}
```
OleksiiBlezniuk, line 68: Now you need to perform extra iterations on every render, instead - prepare array with data after you receive it from server to avoid unnecessary calculations

src/pages/PeoplePage.tsx
```diff
@@ -0,0 +1,72 @@
...
+          )}
+
+          {isLoading ? (
+            <Loader />
+          ) : (
+            <PeopleTable
+              people={people}
+              selectedPerson={personSlug}
+            />
+          )}
```
OleksiiBlezniuk, line 67: Show table only if data is loaded and people array is not empty

OleksiiBlezniuk: Screenshot 2023-04-21 at 16 47 25

OleksiiBlezniuk (changes requested):

PR #668, opened 2023-04-17
react people table basics

andriibezkorovainyi:
[DEMO LINK](https://andriibezkorovainyi.github.io/react_people-table-basics/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,173 +1,77 @@
-import { Loader } from './components/Loader';
-
 import './App.scss';
+import {
+  Navigate, Route, Routes, NavLink,
+} from 'react-router-dom';
+import classNames from 'classnames';
+import { PageNotFound } from './components/PageNotFound/PageNotFound';
+import { HomePage } from './components/HomePage/HomePage';
+import { PeoplePage } from './components/PeoplePage/PeoplePage';
 
-export const App = () => (
-  <div data-cy="app">
-    <nav
-      data-cy="nav"
-      className="navbar is-fixed-top has-shadow"
-      role="navigation"
-      aria-label="main navigation"
-    >
-      <div className="container">
-        <div className="navbar-brand">
-          <a className="navbar-item" href="#/">
-            Home
-          </a>
-
-          <a
-            className="navbar-item has-background-grey-lighter"
-            href="#/people"
-          >
-            People
-          </a>
-        </div>
-      </div>
-    </nav>
-
-    <main className="section">
-      <div className="container">
-        <h1 className="title">Home Page</h1>
-        <h1 className="title">People Page</h1>
-        <h1 className="title">Page not found</h1>
-
-        <div className="block">
-          <div className="box table-container">
-            <Loader />
-
-            <p data-cy="peopleLoadingError" className="has-text-danger">
-              Something went wrong
-            </p>
-
-            <p data-cy="noPeopleMessage">
-              There are no people on the server
-            </p>
-
-            <table
-              data-cy="peopleTable"
-              className="table is-striped is-hoverable is-narrow is-fullwidth"
+export const App = () => {
+  return (
+    <div data-cy="app">
+      <nav
+        data-cy="nav"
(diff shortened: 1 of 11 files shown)


src/App.tsx
```diff
@@ -1,173 +1,77 @@
...
+      >
+        <div className="container">
+          <div className="navbar-brand">
+            <NavLink
+              className={({ isActive }) => classNames(
+                'navbar-item',
+                { 'has-background-grey-lighter': isActive },
+              )}
+              to="/"
             >
```
OleksiiBlezniuk, line 27: Move to separate component

src/components/HomePage/HomePage.tsx
```diff
@@ -0,0 +1,7 @@
+import { FC } from 'react';
+
+export const HomePage: FC = () => {
+  return (
```
OleksiiBlezniuk, line 4:
```suggestion
export const HomePage: FC = () => (
```

src/components/PageNotFound/PageNotFound.tsx
```diff
@@ -0,0 +1,7 @@
+import { FC } from 'react';
+
+export const PageNotFound: FC = () => {
+  return (
```
OleksiiBlezniuk, line 4:
```suggestion
export const PageNotFound: FC = () => (
```

src/components/PeoplePage/Parent/Parent.tsx
```diff
@@ -0,0 +1,24 @@
+import { FC } from 'react';
+import { Person } from '../../../types';
+import { PersonLink } from '../PersonLink/PersonLink';
+import { ParentProps } from '../PeoplePage.types';
+
+function getParentByName(name: string | null, people: Person[]) {
+  const parent = people.find(person => person.name === name);
+
+  return parent || undefined;
```
OleksiiBlezniuk, line 9: But it's already undefined if nothing is found
I think will be more logical to set null, but it's not critical

src/components/PeoplePage/Parent/Parent.tsx
```diff
@@ -0,0 +1,24 @@
...
+}
+
+export const Parent: FC<ParentProps> = ({ parentName, people }) => {
+  const parent = getParentByName(parentName, people);
```
OleksiiBlezniuk, line 13: But now we need to do it on every render, let's prepare array with data after we receive it from server to avoid unnecessary iterations

src/components/PeoplePage/PeoplePage.tsx
```diff
@@ -0,0 +1,61 @@
...
+import { PeopleList } from './PeopleList/PeopleList';
+import { Loader } from '../Loader';
+import { getPeople } from '../../api';
+import { Person } from '../../types';
+
+export const PeoplePage: FC = () => {
+  const [isLoading, setIsLoading] = useState<boolean>(false);
+  const [people, setPeople] = useState<Person[]>([]);
+  const [hasServerError, setHasServerError] = useState(false);
+  const { personSlug = '' } = useParams();
```
OleksiiBlezniuk, line 12: better to create type for params and pass it as generic

src/components/PeoplePage/PeoplePage.tsx
```diff
@@ -0,0 +1,61 @@
...
+            There are no people on the server
+          </p>
+        )}
+
+        {allowPeopleDisplay && (
+          <PeopleList
+            people={people}
+            selectedPersonSlug={personSlug}
+          />
+        )}
```
OleksiiBlezniuk, line 57: Something wrong here, let's show only one error, not both at same time
Screenshot 2023-04-21 at 16 41 35

OleksiiBlezniuk (changes requested):

andriibezkorovainyi:
For some reason test with Linter on the github fails. Local tests are ok. Oleg said thats sometimes it happens and it is normal.

illiarerun (approved):

PR #669, opened 2023-04-17
solution

milenadiamonds1:
[DEMO LINK](https://milenadiamonds1.github.io/react_people-table-basics/)

### DIFF
--- a/src/components/HomePage.tsx
+++ b/src/components/HomePage.tsx
@@ -1,0 +1,7 @@
+export const HomePage = () => {
+  return (
+    <h1 className="title">
+      Home Page
+    </h1>
+  );
+};
--- a/src/components/NavBar.tsx
+++ b/src/components/NavBar.tsx
@@ -1,0 +1,34 @@
+import { NavLink } from 'react-router-dom';
+import classNames from 'classnames';
+
+export const NavBar = () => {
+  return (
+    <nav
+      data-cy="nav"
+      className="navbar is-fixed-top has-shadow"
+      role="navigation"
+      aria-label="main navigation"
+    >
+      <div className="container">
+        <div className="navbar-brand">
+          <NavLink
+            to="/"
+            className={({ isActive }) => classNames('navbar-item', {
+              'has-background-grey-lighter': isActive,
+            })}
+          >
+            Home
+          </NavLink>
+          <NavLink
+            to="people"
+            className={({ isActive }) => classNames('navbar-item', {
+              'has-background-grey-lighter': isActive,
+            })}
+          >
+            People
+          </NavLink>
+        </div>
+      </div>
+    </nav>
+  );
+};
--- a/src/components/PeoplePage.tsx
+++ b/src/components/PeoplePage.tsx
@@ -1,0 +1,62 @@
+import { useState, useEffect } from 'react';
+import { useParams } from 'react-router-dom';
+import { UserList } from './UserList';
+import { Person } from '../types';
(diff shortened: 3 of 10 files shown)


src/components/HomePage.tsx
```diff
@@ -0,0 +1,7 @@
+export const HomePage = () => {
+  return (
```
OleksiiBlezniuk, line 2:
```suggestion
export const HomePage = () => (
```

src/components/NavBar.tsx
```diff
@@ -0,0 +1,34 @@
...
+      aria-label="main navigation"
+    >
+      <div className="container">
+        <div className="navbar-brand">
+          <NavLink
+            to="/"
+            className={({ isActive }) => classNames('navbar-item', {
+              'has-background-grey-lighter': isActive,
+            })}
+          >
```
OleksiiBlezniuk, line 19: Will be nice to move it to separate component

src/components/PeoplePage.tsx
```diff
@@ -0,0 +1,62 @@
...
+import { getPeople } from '../api';
+import { Loader } from './Loader';
+import { ErrorType } from '../Errors';
+
+export const PeoplePage = () => {
+  const [isLoading, setIsLoading] = useState(false);
+  const [users, setUsers] = useState<Person[]>([]);
+  const [errorMessage, setErrorMessage] = useState(false);
+
+  const { slug = '' } = useParams();
```
OleksiiBlezniuk, line 14: Better to add type for params and pass it as generic here

src/components/PeoplePage.tsx
```diff
@@ -0,0 +1,62 @@
...
+      setErrorMessage(true);
+    } finally {
+      setIsLoading(false);
+    }
+  }
+
+  useEffect(() => {
+    setErrorMessage(false);
+    getUsersFromServer();
+    setTimeout(() => setErrorMessage(false), 3000);
```
OleksiiBlezniuk, line 32: I don't think it really useful here, it's more confusing

src/components/PeoplePage.tsx
```diff
@@ -0,0 +1,62 @@
...
+                <p data-cy="peopleLoadingError" className="has-text-danger">
+                  {ErrorType.UnableToLoad}
+                </p>
+              )}
+
+              {users.length <= 0 && (
+                <p data-cy="noPeopleMessage">{ErrorType.NoUsers}</p>
+              )}
+
+              <UserList users={users} personId={slug} />
```
OleksiiBlezniuk, line 55: Better don't render this component if there is some error or there is no data
Also show one error instead of both, it's obvious that there is no users if there were error while data loading

src/components/UserList.tsx
```diff
@@ -0,0 +1,70 @@
...
+          <th>Died</th>
+          <th>Mother</th>
+          <th>Father</th>
+        </tr>
+      </thead>
+
+      <tbody>
+        {users.map((user) => {
+          const mother = findParent(users, user.motherName);
+          const father = findParent(users, user.fatherName);
```
OleksiiBlezniuk, line 38: Map people array and add this fields after you receive people from server, why to do it on every render?

OleksiiBlezniuk (changes requested):
Strange behaviour with error messages, better to not render table if there is no data
https://user-images.githubusercontent.com/84275496/233670026-d728098d-09b8-43ad-b7fa-fde81dfb53ec.mov

PR #672, opened 2023-04-17
Develop

MaksymMaliuk:
[DEMO LINK](https://MaksymMaliuk.github.io/react_people-table-basics/)


src/Pages/NotFoundPage.tsx
```diff
@@ -0,0 +1,9 @@
+import React from 'react';
+
+export const NotFoundPage: React.FC = () => {
+  return (
```
Stepan-TS, line 4: unnecessary return)

src/Pages/HomePage.tsx
```diff
@@ -0,0 +1,9 @@
+import React from 'react';
+
+export const HomePage: React.FC = () => {
+  return (
```
Stepan-TS, line 4: unnecessary return)

src/Pages/PeoplePage.tsx
```diff
@@ -0,0 +1,73 @@
...
+
+      <div className="block">
+        <div className="box table-container">
+          {isError && (
+            <p data-cy="peopleLoadingError" className="has-text-danger">
+              Something went wrong
+            </p>
+          )}
+
+          {(!peopleTable.length && !isLoading) && (
```
Stepan-TS, line 59: it might be better to put this expression into a variable for better readability)

Stepan-TS (approved):
good job, but i suggested fix some code style)

src/Pages/HomePage.tsx
OleksiiBlezniuk, line 4:
```suggestion
export const HomePage: React.FC = () => (
```

src/Pages/PeoplePage.tsx
```diff
@@ -0,0 +1,73 @@
+import React, { useEffect, useState } from 'react';
+import { getPeople } from '../api';
+import { Person } from '../types/Person';
+import { PeopleTable } from '../components/PeopleTable';
+import { Loader } from '../components/Loader';
+
+const findParent = (people: Person[], name: string) => {
+  return people.find(person => person.name === name);
```
OleksiiBlezniuk, line 8: I guess will be more logical to set null if nothing is found, but it's not critical

I mean - undefined tell us that somehow data wasn't set here due to mistake, error or something like that
null - tell us that there shouldn't be anything intentionally

OleksiiBlezniuk (changes requested):
Let's render table only when there is no errors
Also let's show only one error message at one time, if there was error while loading - show error about it, not about array is empty
Screenshot 2023-04-19 at 12 14 54

OleksiiBlezniuk (approved):
GJ!

PR #683, opened 2023-04-20
solics

vadosik15:
[DEMO LINK](https://vadosik15.github.io/react_people-table-basics/)

### DIFF
--- a/src/components/Navigation/Navigation.tsx
+++ b/src/components/Navigation/Navigation.tsx
@@ -1,0 +1,20 @@
+import { PageNavLink } from '../PageNavLink/PageNavLink';
+
+export const Navigation = () => {
+  return (
+    <nav
+      data-cy="nav"
+      className="navbar is-fixed-top has-shadow"
+      role="navigation"
+      aria-label="main navigation"
+    >
+      <div className="container">
+        <div className="navbar-brand">
+          <PageNavLink to="/" title="Home" />
+
+          <PageNavLink to="/people" title="People" />
+        </div>
+      </div>
+    </nav>
+  );
+};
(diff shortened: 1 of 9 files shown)


src/components/PersonRow/PersonRow.tsx
```diff
@@ -0,0 +1,50 @@
...
+    motherName,
+    slug,
+    mother,
+    father,
+  } = person;
+
+  return (
+    <tr
+      data-cy="person"
+      key={slug}
```
OleksiiBlezniuk, line 26: The 'key' attribute should not be used on a 'tr' element. It should be applied to the parent component when mapping over an array of data. The 'key' attribute can be removed in this case. 🔑

src/components/PeoplePage/PeoplePage.tsx
```diff
@@ -0,0 +1,79 @@
...
+import { Person } from '../../types';
+import { getPeople } from '../../api';
+import { PeopleTable } from '../PeopleTable/PeopleTable';
+import { Loader } from '../Loader';
+
+export const PeoplePage = () => {
+  const [people, setPeople] = useState<Person[]>([]);
+  const [errorMessage, setErrorMessage] = useState('');
+  const [isLoading, setIsLoading] = useState(false);
+  const { slug = '' } = useParams();
```
OleksiiBlezniuk, line 12: You should provide a type for useParams, like `useParams<{ slug: string }>()` to ensure type safety for the slug parameter.

src/components/Navigation/Navigation.tsx
```diff
@@ -0,0 +1,20 @@
+import { PageNavLink } from '../PageNavLink/PageNavLink';
+
+export const Navigation = () => {
+  return (
```
OleksiiBlezniuk, line 4:
```suggestion
export const Navigation = () => (
```

OleksiiBlezniuk (approved):

PR #688, opened 2023-04-24
Solution

katerinalex:
[DEMO LINK](https://katerinalex.github.io/react_people-table-basics/)


IvanFesenko (changes requested):
Please check comment.
It should work same as working example https://mate-academy.github.io/react_people-table-basics/#/people

PR #698, opened 2023-05-02
solution

illia-kyselov:
[DEMO LINK](https://illia-kyselov.github.io/react_people-table-basics/)


src/components/Loader/Person/Person.tsx
```diff
@@ -0,0 +1,21 @@
...
+import { Person } from '../../../types/Person';
+
+type Props = {
+  person: Person;
+};
+
+export const PersonLink: React.FC<Props> = ({ person }) => (
+  <td>
+    <Link
+      className={classNames('', {
```
vlad-demchuk, line 13: The 'className' attribute is assigned an empty string as the first argument in 'classNames'. You can remove the empty string if it's not necessary.

src/App.tsx
```diff
@@ -1,172 +1,23 @@
-import { Loader } from './components/Loader';
-
+import { Navigate, Route, Routes } from 'react-router-dom';
 import './App.scss';
+import { Navigation } from './components/Loader/Navigation/Navigation';
```
vlad-demchuk, line 3: why do you have Navigation components inside Loader folder?

src/App.tsx
```diff
@@ -1,172 +1,23 @@
...
+import { People } from './components/Loader/People/People';
```
vlad-demchuk, line 4: the same

src/App.tsx
```diff
@@ -1,172 +1,23 @@
...
-                </tr>
-              </tbody>
-            </table>
-          </div>
-        </div>
+        <Routes>
+          <Route path="*" element={<h1 className="title">Page not found</h1>} />
+          <Route path="/" element={<h1 className="title">Home Page</h1>} />
+          <Route path="home" element={<Navigate to="/" replace />} />
+          <Route path="/people">
```
vlad-demchuk, line 16: Use a leading slash for the 'home' path. It should be '/home' instead of 'home'

src/components/Loader/People/People.tsx
```diff
@@ -0,0 +1,136 @@
...
+                        <th>Mother</th>
+                        <th>Father</th>
+                      </tr>
+                    </thead>
+
+                    <tbody>
+                      {listPeople.map((person: Person) => (
+                        <tr
+                          data-cy="person"
+                          className={classNames('', {
```
vlad-demchuk, line 87: Remove the empty string from the `classNames` function call, as it is not necessary.

src/components/Loader/People/People.tsx
```diff
@@ -0,0 +1,136 @@
...
+
+  const getParent = (person: Person) => {
+    return listPeople
+      .find(
+        (el: Person) => el.name === person.fatherName,
+      );
+  };
+
+  const isActive = (name: string) => {
+    return slug.replaceAll('-', ' ').toLowerCase().includes(name.toLowerCase());
```
vlad-demchuk, line 28: why do we need this?

src/components/Loader/People/People.tsx
```diff
@@ -0,0 +1,136 @@
...
+  const emptyName = (person: Person) => {
+    return person.motherName || '-';
+  };
+
+  const createSlug = (name: string) => {
+    const parent = listPeople.filter((el: Person) => el.name === name)[0];
+
+    return parent
+      ? `/people/${parent.name.replaceAll(' ', '-')}-${parent.born}`
+      : '';
```
vlad-demchuk, line 40: the same here

why do we need to change slug?

vlad-demchuk (changes requested):
fix changing styles for selected people
https://user-images.githubusercontent.com/102411130/235845979-906390b3-8d48-44e0-9bb7-f74511e340c0.mov

AlisaVlasova (approved):

PR #712, opened 2023-05-16
people table is ready

PavloReutskiy:
[DEMO LINK](https://PavloReutskiy.github.io/react_people-table-basics/)


PavloReutskiy:
I can't figure out what is the problem with tests. The only thing that was changed is timeout for cypress. There is problem with eslint-plugin but all tests were passed local

image

image

src/api.ts
```diff
@@ -1,15 +0,0 @@
-import { Person } from './types/Person';
```
jstmpelowycz, line 1: I think you could've kept this code separately, then use it in your custom hook

src/custom-hooks/useFetch.ts
```diff
@@ -0,0 +1,31 @@
+import { useEffect, useState } from 'react';
+import { Person } from '../types/Person';
+
+function wait(delay: number) {
+  return new Promise(resolve => setTimeout(resolve, delay));
+}
+
+export function useFetch(url: string) {
```
jstmpelowycz, line 8:
```suggestion
export function userFetchPeople() {
```
this hook is tied to people-related flow, so there's no need to pass url as a param — we always retrieve users

src/pages/PeoplePage/PeoplePage.tsx
```diff
@@ -0,0 +1,51 @@
+import { FC } from 'react';
+import { useFetch } from '../../custom-hooks/useFetch';
+import { Loader } from '../../components/Loader';
+import { PeopleTable } from '../../components/PeopleTable';
+import { API_URL } from '../../constants/apiUrl';
+
+export const PeoplePage: FC = () => {
+  const { people, isLoading, errorMessage } = useFetch(API_URL);
```
jstmpelowycz, line 8: this is a really good approach, mostly it works this way in real project

src/pages/PeoplePage/PeoplePage.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+  const getPerent = (parentName: string | null) => {
```
jstmpelowycz, line 10:
```suggestion
  const findParent = (parentName: string | null) => {
```

src/pages/PeoplePage/PeoplePage.tsx
```diff
@@ -0,0 +1,51 @@
...
+    if (parentName) {
+      return people?.find(person => person.name === parentName);
+    }
+
+    return null;
+  };
+
+  const peopleWithPerents = people ? people.map((person) => {
```
jstmpelowycz, line 18:
```suggestion
  const peopleWithParents = people ? people.map((person) => {
```

jstmpelowycz: do you really need this ternary? seems like it's a bit complicated here
in your custom hook, just return empty array to avoid this validation

jstmpelowycz (approved):

PavloReutskiy:
all done! thx a lot, now it looks much better

src/pages/PeoplePage/PeoplePage.tsx
jstmpelowycz: great research, keep on!

PR #717, opened 2023-05-17
completed

Zad0rnik:
[DEMO LINK](https://Zad0rnik.github.io/react_people-table-basics/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -14,15 +14,8 @@ export const App = () => (
         <Routes>
           <Route path="/" element={<HomePage />} />
           <Route path="/people" element={<PeoplePage />}>
-            <Route
-              path=":selectedPerson"
-              element={<PeoplePage />}
-            />
-
-            <Route
-              index
-              element={<PeoplePage />}
-            />
+            <Route path=":selectedPerson" element={<PeoplePage />} />
+            <Route index element={<PeoplePage />} />
           </Route>
           <Route path="/home" element={<Navigate to="/" replace />} />
           <Route path="*" element={<NotFoundPage />} />
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,172 +1,32 @@
-import { Loader } from './components/Loader';
-
 import './App.scss';
+import { Routes, Route, Navigate } from 'react-router-dom';
+import { HomePage } from './components/HomePage';
+import { NavBar } from './components/NavBar';
+import { NotFoundPage } from './components/NotFoundPage';
+import { PeoplePage } from './components/PeoplePage';
 
 export const App = () => (
   <div data-cy="app">
-    <nav
-      data-cy="nav"
-      className="navbar is-fixed-top has-shadow"
-      role="navigation"
-      aria-label="main navigation"
-    >
-      <div className="container">
-        <div className="navbar-brand">
-          <a className="navbar-item" href="#/">
-            Home
-          </a>
-
-          <a
-            className="navbar-item has-background-grey-lighter"
-            href="#/people"
-          >
-            People
-          </a>
-        </div>
-      </div>
-    </nav>
+    <NavBar />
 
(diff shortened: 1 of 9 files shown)


src/components/NavBar.tsx
```diff
@@ -0,0 +1,18 @@
+import { memo } from 'react';
+import { PageNavLink } from './PageNavLink';
+
+export const NavBar = memo(() => (
```
jstmpelowycz, line 4: > since the HomePage component appears to be a simple static component without any dynamic behavior, memoizing it may not provide any significant benefits

src/components/PeopleList.tsx
```diff
@@ -0,0 +1,40 @@
...
+          <th>Father</th>
+        </tr>
+      </thead>
+
+      <tbody>
+        {people.map((person) => (
+          <PersonCard
+            person={person}
+            key={person.slug}
+            selectedPerson={selectedPerson}
```
jstmpelowycz, line 34: seems like you can use `useParams` down in the `PersonLink`

src/components/PeopleList.tsx
```diff
@@ -0,0 +1,40 @@
...
+          <th>Born</th>
+          <th>Died</th>
+          <th>Mother</th>
+          <th>Father</th>
+        </tr>
+      </thead>
+
+      <tbody>
+        {people.map((person) => (
+          <PersonCard
```
jstmpelowycz, line 31: it's rather link, not card

src/components/PeoplePage.tsx
```diff
@@ -0,0 +1,72 @@
...
+      const peopleFromServer = await getPeople();
+
+      const peopleWithParents = peopleFromServer.map((person) => ({
+        ...person,
+        mother: findParent(peopleFromServer, person.motherName),
+        father: findParent(peopleFromServer, person.fatherName),
+      }));
+
+      setPeople(peopleWithParents);
+      setIsLoading(false);
```
jstmpelowycz, line 31: disable loading in `finally` block

src/components/PeoplePage.tsx
```diff
@@ -0,0 +1,72 @@
...
+const findParent = (people: Person[], parentName: string | null) => {
+  return people.find((person) => person.name === parentName);
+};
+
+export const PeoplePage = memo(() => {
+  const [people, setPeople] = useState<Person[]>([]);
+  const [isLoading, setIsLoading] = useState(false);
+  const [errorMessage, setErrorMessage] = useState<string | null>(null);
+
+  const LoadPeople = useCallback(async () => {
```
jstmpelowycz, line 18:
```suggestion
  const loadPeople = useCallback(async () => {
```

jstmpelowycz (changes requested):

src/App.tsx
```diff
@@ -1,172 +1,32 @@
...
+          <Route path="/people" element={<PeoplePage />}>
+            <Route
+              path=":selectedPerson"
+              element={<PeoplePage />}
+            />
+
+            <Route
+              index
+              element={<PeoplePage />}
+            />
```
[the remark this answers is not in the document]
Zad0rnik: Снимок экрана 2023-05-23 в 16 03 59

if i remove element from parent route it will cause this problem

src/App.tsx
```diff
@@ -1,172 +1,25 @@
...
-                    </a>
-                  </td>
-                </tr>
-              </tbody>
-            </table>
-          </div>
-        </div>
+        <Routes>
+          <Route path="/" element={<HomePage />} />
+          <Route path="/people" element={<PeoplePage />}>
```
jstmpelowycz, line 16: afaik, it is not necessary to pass the element prop to the parent route if you want to render different components for each route

jstmpelowycz (approved):

PR #724, opened 2023-05-18
solution

Olekk17:
[DEMO LINK](https://olekk17.github.io/react_people-table-basics/)

### DIFF
--- a/src/components/PeoplePage/PeoplePage.tsx
+++ b/src/components/PeoplePage/PeoplePage.tsx
@@ -19,7 +19,7 @@ export const PeoplePage: React.FC = () => {
   const [people, setPeople] = useState<Person[]>([]);
   const [isLoading, setIsLoading] = useState(false);
   const [isError, setIsError] = useState(false);
-  const isListEmpty = !people.length;
+  const isListEmpty = !people.length && !isLoading && !isError;
 
   const loadPeople = useCallback(async () => {
     setIsLoading(true);


src/App.tsx
```diff
@@ -12,161 +15,30 @@ export const App = () => (
     >
```
SerhiiKirik, line 15: Move `nav` to the separate component

src/App.tsx
```diff
@@ -12,161 +15,30 @@ export const App = () => (
...
 
-                  <td>
-                    <a href="#/people/emile-haverbeke-1877">
-                      Emile Haverbeke
-                    </a>
-                  </td>
-                </tr>
+          <Route path="people">
+            <Route index element={<PeoplePage />} />
+            <Route path=":selectedPersonSlug" element={<PeoplePage />} />
```
SerhiiKirik, line 34: you can just use `slug` here
```suggestion
            <Route path=":slug" element={<PeoplePage />} />
```

src/components/PeoplePage/PeoplePage.tsx
```diff
@@ -0,0 +1,88 @@
...
+      setPeople(preparedPeople);
+
+      if (!peopleFromServer.length) {
+        setIsListEmpty(true);
+      }
+    } catch {
+      setIsError(true);
+    }
+
+    setIsLoading(false);
```
SerhiiKirik, line 53:
```suggestion
    } catch {
      setIsError(true);
    } finally {
      setIsLoading(false);
    }
```

src/components/PeoplePage/PeoplePage.tsx
```diff
@@ -0,0 +1,88 @@
...
+          peopleFromServer,
+          person.fatherName,
+        );
+
+        return personCopy;
+      });
+
+      setPeople(preparedPeople);
+
+      if (!peopleFromServer.length) {
```
SerhiiKirik, line 46: why do you need it?
You can handle it without extra state

SerhiiKirik (changes requested):
Nice 🔥 
Let's make a few improvements 🧑‍💻

src/components/PeoplePage/PeoplePage.tsx
Olekk17: I wanted not to show this message when people are loading from the server for the first time, because before that it is always an empty array and this message will be shown when people are loading

SerhiiKirik: But you have put a two check `...length && !loading`

SerhiiKirik (approved):
Nice 🔥

PR #730, opened 2023-05-18
Solution

Maks-Tsarenko:
[DEMO LINK](https://Maks-Tsarenko.github.io/react_people-table-basics/)


src/utils/findParents.tsx
anstsot, line 1: it will be better to rename this file to be a component `ParentLink`

anstsot (approved):
Good job, approved

PR #744, opened 2023-05-22
Solution

DimaDvm:
DEMO LINK - https://DimaDvm.github.io/react_people-table-basics/


jstmpelowycz (changes requested):

src/components/HomePage.tsx
```diff
@@ -0,0 +1,4 @@
+/* eslint-disable linebreak-style */
```
jstmpelowycz, line 1: why do you need this comment? if it's all about LF/CRLF, talk to @glots-a

jstmpelowycz (approved):

PR #754, opened 2023-05-24
all works

BondarenkoDmitriy:
[DEMO LINK](https://BondarenkoDmitriy.github.io/react_people-table-basics/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,173 +1,27 @@
-import { Loader } from './components/Loader';
-
+import { Navigate, Route, Routes } from 'react-router-dom';
 import './App.scss';
+import { HomePage } from './components/HomePage';
+import { PeoplePage } from './components/PeoplePage';
+import { NotFoundPage } from './components/NotFoundPage';
+import { Navbar } from './components/NavigationBar';
 
 export const App = () => (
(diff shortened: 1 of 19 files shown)


src/components/Loader/Loader.tsx
```diff
@@ -1,7 +1,8 @@
+import { memo } from 'react';
 import './Loader.scss';
 
-export const Loader = () => (
+export const Loader = memo(() => (
```
SerhiiKirik, line 4: why odo you need `memo` if you don't have a props?

SerhiiKirik (changes requested):
Well done 🔥 
Let's fix some problems 🧑‍💻 
You delete some wrapper
image
I can't select a user after i already chose someone

image

src/App.tsx
```diff
@@ -1,173 +1,27 @@
...
-          >
-            People
-          </a>
-        </div>
-      </div>
-    </nav>
+    <Navbar />
 
     <main className="section">
-      <div className="container">
```
SerhiiKirik, line 30: don't remove this container

SerhiiKirik (approved):
Good job 🔥 
Still not fixed
image

PR #772, opened 2023-05-31
Solution

Zibi95:
[DEMO LINK](https://Zibi95.github.io/react_people-table-basics/)


Radoslaw-Czerniawski (approved):
Everything fine, posted a comment with a proposal for custom hook `useFetch`, might be a good idea to try coding something like that and digging deeper to grasp how it works :)

PR #805, opened 2023-06-22
Develop

generatedman:
[DEMO LINK](https://generatedman.github.io/react_people-table-basics/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,8 @@
+import { Routes, Route, NavLink } from 'react-router-dom';
 import { Loader } from './components/Loader';
+import { PeoplePage } from './PeoplePage';
+import { Navigate } from 'react-router-dom';
+import classNames from 'classnames';
 
 import './App.scss';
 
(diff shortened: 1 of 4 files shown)


src/PeoplePage.tsx
```diff
@@ -0,0 +1,64 @@
...
+                <th>Born</th>
+                <th>Died</th>
+                <th>Mother</th>
+                <th>Father</th>
+              </tr>
+            </thead>
+
+            <tbody>
+              {people.map(person => (
+                <PersonLink person={person} people={people} />
```
roman-mirzoian, line 57: The `PersonLink` component is missing a 'key' prop. In React, a 'key' is necessary when rendering a list of children components.

src/App.tsx
```diff
@@ -12,161 +19,53 @@ export const App = () => (
     >
       <div className="container">
         <div className="navbar-brand">
-          <a className="navbar-item" href="#/">
+          <NavLink
+            className={
```
roman-mirzoian, line 23: It is preferable to move the function that generates classes outside the render method to avoid creating a new function on each render.
This applies to all code.

src/PersonLink.tsx
```diff
@@ -0,0 +1,79 @@
+import React from 'react';
+import classNames from 'classnames';
+import { useParams, Link } from 'react-router-dom';
+import { Person } from './types';
+
+const findParent = (
```
roman-mirzoian, line 6: The name of the `findParent` function does not fully convey the meaning of this function, because in addition to the search, we also return a reference to the parent, if it exists. Therefore, it would be better to name this function something like `getParentLink`.

src/PersonLink.tsx
```diff
@@ -0,0 +1,79 @@
...
+  person: Person,
+  parentType: string,
+  name: string,
+  peopleList: Person[],
```
roman-mirzoian, line 10: We can make typing look a little better by bringing the types into a separate interface or type.

src/PersonLink.tsx
```diff
@@ -0,0 +1,79 @@
+import React from 'react';
+import classNames from 'classnames';
+import { useParams, Link } from 'react-router-dom';
+import { Person } from './types';
+
```
roman-mirzoian, line 5: Before starting to work with functions and components, it would be very cool to define the variables that we use further into separate constants, so that we do not use magic values:
```
const FATHER = 'father';
const MOTHER = 'mother';
const FEMALE = 'f';
const NOT_SET_VALUE = '-';
```
This will provide us with better readability of the code itself, conditions in the code and understanding that we will not accidentally make a mistake in a line.

src/PersonLink.tsx
```diff
@@ -0,0 +1,79 @@
...
+  if (parent) {
+    return (
+      <Link
+        className={classNames(
+          { 'has-text-danger': parentType === 'mother' },
+        )}
+        to={`/people/${parent.slug}`}
+      >
+        {parentType === 'mother'
```
roman-mirzoian, line 23: It is better to put such constructions into separate variables and substitute the already calculated result into the component.

src/PersonLink.tsx
```diff
@@ -0,0 +1,79 @@
...
+  name: string,
+  peopleList: Person[],
+) => {
+  const parent = peopleList
+    .find(personToFind => personToFind.name === name);
+
+  if (parent) {
+    return (
+      <Link
+        className={classNames(
```
roman-mirzoian, line 18: To improve the readability of the code, it is better to transfer the definition of classes to separate variables and use the ready-made result. You can see an example of use on the website of the [classnames ](https://www.npmjs.com/package/classnames) package.
This applies to all code.

src/PersonLink.tsx
```diff
@@ -0,0 +1,79 @@
...
+          )}
+          to={`/people/${person.slug}`}
+        >
+          {person.name}
+        </Link>
+      </td>
+
+      <td>{person.sex}</td>
+      <td>{person.born}</td>
+      <td>{person.died}</td>
```
roman-mirzoian, line 62: `person` parameters are now passed correctly, but it can be done in a more modern way by destructuring the `person` object: `const {sex, born, etc.} = person;`

src/PeoplePage.tsx
```diff
@@ -0,0 +1,64 @@
...
+            className="table is-striped is-hoverable is-narrow is-fullwidth"
+          >
+            <thead>
+              <tr>
+                <th>Name</th>
+                <th>Sex</th>
+                <th>Born</th>
+                <th>Died</th>
+                <th>Mother</th>
+                <th>Father</th>
```
roman-mirzoian, line 51: Here you can cheat a little and make the code more readable and scalable if we take out the names of the columns in a separate array and map it as we do with `people`.

src/PeoplePage.tsx
```diff
@@ -0,0 +1,64 @@
...
+      {isLoading && <Loader />}
+
+      {error
+        && (
+          <p data-cy="peopleLoadingError" className="has-text-danger">
+            Something went wrong
+          </p>
+        )}
+
+      {(!error && !isLoading)
```
roman-mirzoian, line 38: If you need to specify more than one condition in the component itself, it is better to put it in a separate variable and name it so that it combines this condition in its name (For example: `isLoadingSuccessful`).

roman-mirzoian (changes requested):
A lot of good work has been done, some optimization remains.

src/App.tsx
```diff
@@ -12,161 +19,53 @@ export const App = () => (
...
+              ({ isActive }) => classNames('navbar-item',
+                {
+                  'has-background-grey-lighter': isActive,
+                })
+            }
+            to="/people"
           >
             People
-          </a>
+          </NavLink>
```
maksym-bobryk, line 44: You can also create a separated component for avoiding code duplication by custom NavLink.

maksym-bobryk (approved):

src/App.tsx
```diff
@@ -12,161 +19,53 @@ export const App = () => (
...
+            element={<PeoplePage />}
+          />
+          <Route
+            path="/people/:slug"
+            element={<PeoplePage />}
+          />
+
+          <Route
+            path="*"
+            element={<h1 className="title">Page not found</h1>}
```
dinaminahi, line 66: Let's create a component for this

src/PersonLink.tsx
```diff
@@ -0,0 +1,109 @@
...
+
+interface ParentLink {
+  person: Person,
+  parentType: string,
+  name: string,
+  peopleList: Person[],
+}
+
+const getParentLink = (
+  args: ParentLink,
```
dinaminahi, line 19:
```suggestion
  props: ParentLink,
```
We usually name it `props`

src/App.tsx
```diff
@@ -12,161 +19,53 @@ export const App = () => (
     >
       <div className="container">
         <div className="navbar-brand">
```
dinaminahi, line 21: Put nav to a Navigation component and create a Layout component that will serve as a wrapper for all your routes. Within this Layout, we'll include a Navigation component and a main section to wrap the content. To achieve this, we can use the Outlet component from react-router-dom.

```
<Routes>
  <Route path="/" element={<Layout />}>
    .. homepage
    ...peoplepage
    ...
 </Routes>   
```

dinaminahi (changes requested):
Almost done 👍 
Just few fixes needed

PR #812, opened 2023-07-02
Add task solution

IrynaBalandiukh:
[DEMO LINK](https://IrynaBalandiukh.github.io/react_people-table-basics/)

### DIFF
--- a/src/components/PeopleTable/PeopleTable.tsx
+++ b/src/components/PeopleTable/PeopleTable.tsx
@@ -1,0 +1,41 @@
+import React from 'react';
+import { useParams } from 'react-router-dom';
+
+import { Person } from '../../types';
+import { PersonInfo } from '../PersonInfo';
+
+type Props = {
+  people: Person[];
+};
+
+export const PeopleTable: React.FC<Props> = ({ people }) => {
+  const { slug = '' } = useParams();
+
+  return (
+    <table
+      data-cy="peopleTable"
+      className="table is-striped is-hoverable is-narrow is-fullwidth"
+    >
+      <thead>
+        <tr>
+          <th>Name</th>
+          <th>Sex</th>
+          <th>Born</th>
+          <th>Died</th>
+          <th>Mother</th>
+          <th>Father</th>
+        </tr>
+      </thead>
+
+      <tbody>
+        {people.map(person => (
+          <PersonInfo
+            key={person.slug}
+            person={person}
+            selectedPersonSlug={slug}
+          />
+        ))}
+      </tbody>
+    </table>
+  );
+};
--- a/src/components/PersonLink/PersonLink.tsx
+++ b/src/components/PersonLink/PersonLink.tsx
@@ -1,0 +1,19 @@
+import React from 'react';
+import { NavLink } from 'react-router-dom';
+import cn from 'classnames';
+
+import { Person } from '../../types';
+
+type Props = {
+  person: Person;
+};
+
+export const PersonLink: React.FC<Props> = ({ person }) => {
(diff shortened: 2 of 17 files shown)


src/pages/PeoplePage/PeoplePage.tsx
```diff
@@ -0,0 +1,65 @@
...
+  const [isError, setIsError] = useState(false);
+
+  useEffect(() => {
+    getPeople()
+      .then(response => setPeople(response))
+      .catch(() => setIsError(true))
+      .finally(() => setIsLoaded(true));
+  }, []);
+
+  const preparedPeople: Person[] = people.map(person => {
```
roman-mirzoian, line 20: The right decision, but now you have two searches in the middle of another search, which is algorithmically not optimal.
1. To improve the algorithm, you can create a separate map (`peopleMap`) and add all the people to it using their name as a key (`peopleMap.set(person.name, person)`), after that go through the `people` array and find the person's father and mother (`peopleMap.get(person.motherName)`, `peopleMap.get(person.fatherName)`).
This is a more complicated approach, but more correct on real projects.
2. For code cleanliness, this function could be moved outside the component and used as a helper.
3. Now, with each rendering of the component, you will search and prepare the `preparedPeople` array - this can be improved by calling this search in the `useEffect` hook.

src/components/PersonLink/PersonLink.tsx
```diff
@@ -0,0 +1,19 @@
...
+
+type Props = {
+  person: Person;
+};
+
+export const PersonLink: React.FC<Props> = ({ person }) => {
+  const { slug, name, sex } = person;
+
+  return (
+    <NavLink to={`/people/${slug}`} className={cn({ 'has-text-danger': sex === 'f' })}>
```
roman-mirzoian, line 15: It would be cool to follow suit not to use magic values in the code, but to declare them through variables:
```
const FEMALE = 'f';
```
Now it is not necessary to fix it, but in real projects people try to write it like this. This may seem insignificant, but on large projects this approach can save you from mistakes.

src/components/PeopleTable/PeopleTable.tsx
```diff
@@ -0,0 +1,41 @@
...
+      className="table is-striped is-hoverable is-narrow is-fullwidth"
+    >
+      <thead>
+        <tr>
+          <th>Name</th>
+          <th>Sex</th>
+          <th>Born</th>
+          <th>Died</th>
+          <th>Mother</th>
+          <th>Father</th>
```
roman-mirzoian, line 26: Here we can cheat a little and improve the code by declaring an array with all the column names and looping through that array just like we currently do with the `people` array.

roman-mirzoian (changes requested):
Well done and you did a good job, but we can improve the performance of the code a bit :)

snuber-w (approved):

src/components/PeopleTable/PeopleTable.tsx
```diff
@@ -0,0 +1,38 @@
...
+  const { slug = '' } = useParams();
+
+  return (
+    <table
+      data-cy="peopleTable"
+      className="table is-striped is-hoverable is-narrow is-fullwidth"
+    >
+      <thead>
+        <tr>
+          {['Name', 'Sex', 'Born', 'Died', 'Mother', 'Father'].map(column => (
```
Moroz-Dmytro, line 21: 👍 but it would be better to create a `const` value for this array and then use this value here

Moroz-Dmytro (approved):
LGTM 👍 
Only small advice

PR #818, opened 2023-07-06
Solution

BogdanParubchenk01:
[DEMO LINK](https://BogdanParubchenk01.github.io/react_people-table-basics/)

### DIFF
--- a/src/CustomHooks/usePeopleData/usePeopleData.tsx
+++ b/src/CustomHooks/usePeopleData/usePeopleData.tsx
@@ -1,0 +1,66 @@
+import { useEffect, useState } from 'react';
+import { Person } from '../../types';
+import { getPeople } from '../../api';
+import { TableList } from '../../components/TableList';
+import { Loader } from '../../components/Loader';
+
+interface UsePeopleDataResult {
+  content: JSX.Element | null;
+}
+
+function sortPersons(persons: Person[]) {
+  const valuePersons = [...persons];
+
+  return valuePersons
+    .map(person => {
+      return {
+        ...person,
+        mother: valuePersons.find(mom => mom.name === person.motherName),
+        father: valuePersons.find(dad => dad.name === person.fatherName),
+      };
+    });
+}
+
+export const usePeopleData = (): UsePeopleDataResult => {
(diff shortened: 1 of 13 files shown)


src/CustomHooks/usePeopleData/usePeopleData.tsx
```diff
@@ -0,0 +1,66 @@
...
+      return {
+        ...person,
+        mother: valuePersons.find(mom => mom.name === person.motherName),
+        father: valuePersons.find(dad => dad.name === person.fatherName),
+      };
+    });
+}
+
+export const usePeopleData = (): UsePeopleDataResult => {
+  const [dataPerosn, setDataPerson] = useState<Person[]>([]);
```
VitaliyBondarenko1982, line 25:
```suggestion
  const [dataPerson, setDataPerson] = useState<Person[]>([]);
```
typo

src/CustomHooks/usePeopleData/usePeopleData.tsx
```diff
@@ -0,0 +1,66 @@
...
+    );
+  } else if (dataPerosn.length) {
+    content = <TableList persons={dataPerosn} />;
+  } else {
+    content = (
+      <p data-cy="noPeopleMessage">
+        There are no people on the server
+      </p>
+    );
+  }
```
VitaliyBondarenko1982, line 63: consider to wrap this logic in useMemo and avoid else if statements for better readibility.
will be something like that.
looks like in any case it will be JSX.Element (not null)

```suggestion
  let content: JSX.Element = useMemo(() => {
    if (isLoading) {
    return <Loader />;
  }
  
   if (loadError) {
    return  (
      <p data-cy="peopleLoadingError" className="has-text-danger">
        Something went wrong
      </p>
    );
  } 
  
  if (dataPerson.length) {
    return <TableList persons={dataPerosn} />;
  }
  
    return (
      <p data-cy="noPeopleMessage">
        There are no people on the server
      </p>
    );
  }, [isLoading, loadError, dataPerson.length]);
  
```

VitaliyBondarenko1982 (changes requested):

VitaliyBondarenko1982 (approved):

PR #824, opened 2023-07-08
solution

annholovko:
[DEMO LINK](https://annholovko.github.io/react_people-table-basics/)

### DIFF
--- a/src/components/PeopleTable/PeopleTable.tsx
+++ b/src/components/PeopleTable/PeopleTable.tsx
@@ -1,0 +1,34 @@
+import { FC } from 'react';
+import { Person } from '../../types';
+import { PersonInfo } from '../PersonInfo';
+
+interface Props {
+  people: Person[]
+}
+export const PeopleTable: FC<Props> = ({ people }) => {
+  return (
+    <table
+      data-cy="peopleTable"
+      className="table is-striped is-hoverable is-narrow is-fullwidth"
+    >
+      <thead>
+        <tr>
+          <th>Name</th>
+          <th>Sex</th>
+          <th>Born</th>
+          <th>Died</th>
+          <th>Mother</th>
+          <th>Father</th>
+        </tr>
+      </thead>
+
+      {people.map((person) => (
+        <PersonInfo
+          data-cy="person"
+          key={person.slug}
+          person={person}
+        />
+      ))}
+    </table>
+  );
+};
--- a/src/components/PersonInfo/PersonInfo.tsx
+++ b/src/components/PersonInfo/PersonInfo.tsx
@@ -1,0 +1,40 @@
+import { FC } from 'react';
+import { useParams } from 'react-router-dom';
+import cn from 'classnames';
+import { Person } from '../../types';
+import { PersonLink } from '../PersonLink';
+
+interface Props {
+  person: Person
+}
+export const PersonInfo: FC<Props> = ({ person }) => {
+  const { slug } = useParams<{ slug: string }>();
+
+  return (
+    <tbody>
+      <tr
+        data-cy="person"
+        className={cn({ 'has-background-warning': person.slug === slug })}
+      >
+        <td>
+          <PersonLink person={person} />
+        </td>
+
+        <td>{person.sex}</td>
+        <td>{person.born}</td>
+        <td>{person.died}</td>
+        <td>
+          {person.mother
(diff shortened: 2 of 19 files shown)


src/components/PersonLink/PersonLink.tsx
```diff
@@ -0,0 +1,19 @@
...
+import { Person } from '../../types';
+
+interface Props {
+  person: Person
+}
+
+export const PersonLink: FC<Props> = ({ person }) => (
+  <Link
+    className={cn({
+      'has-text-danger': person.sex === 'f',
```
roman-mirzoian, line 13: It would be cool not to use magic values and put such strings into constants and reuse them:
```
const FEMALE = 'f';
const NOT_SET_VALUE = '-';
```
This will increase the readability of the code and save us from possible errors.
This applies to development in general.

src/pages/PeoplePage/PeoplePage.tsx
```diff
@@ -0,0 +1,73 @@
...
+        setIsLoading(false);
+      }
+    };
+
+    fetchData();
+  }, []);
+
+  const visiblePeople: Person[] = people.map(person => {
+    const mother = people.find(m => m.name === person.motherName);
+    const father = people.find(f => f.name === person.fatherName);
```
roman-mirzoian, line 30: The right decision, but now you have two searches in the middle of another search, which is algorithmically not optimal.
To improve the algorithm, you can create a separate map (`peopleMap`) and add all the people to it using their name as a key (`peopleMap.set(person.name, person)`), after that go through the `people` array and find the person's father and mother (`peopleMap.get(person.motherName)`, `peopleMap.get(person.fatherName)`).
This is a more complicated approach, but more correct on real projects.

src/pages/PeoplePage/PeoplePage.tsx
```diff
@@ -0,0 +1,73 @@
...
+        setIsError(true);
+      } finally {
+        setIsLoading(false);
+      }
+    };
+
+    fetchData();
+  }, []);
+
+  const visiblePeople: Person[] = people.map(person => {
```
roman-mirzoian, line 28: Searching for the neede people can be carried out as a separate helper outside the component, this will increase the productivity and readability of the code.

src/pages/PeoplePage/PeoplePage.tsx
```diff
@@ -0,0 +1,73 @@
...
+  const [isError, setIsError] = useState(false);
+  const [isLoading, setIsLoading] = useState(true);
+
+  useEffect(() => {
+    const fetchData = async () => {
+      try {
+        const peopleFromServer = await getPeople();
+
+        setPeople(peopleFromServer);
+      } catch {
```
roman-mirzoian, line 18: In the `catch` block, you should also catch a possible `error`, because now the error does not look informative, for example, you can do this:
```
catch(error) { setIsError(`Something went wrong: ${error.message}`)}
```

src/components/PersonInfo/PersonInfo.tsx
```diff
@@ -0,0 +1,40 @@
...
+        data-cy="person"
+        className={cn({ 'has-background-warning': person.slug === slug })}
+      >
+        <td>
+          <PersonLink person={person} />
+        </td>
+
+        <td>{person.sex}</td>
+        <td>{person.born}</td>
+        <td>{person.died}</td>
```
roman-mirzoian, line 25: It would be good to use destructuring in such places:
```
const { sex, born, etc. } = person;
```

src/components/PeopleTable/PeopleTable.tsx
```diff
@@ -0,0 +1,34 @@
...
+      className="table is-striped is-hoverable is-narrow is-fullwidth"
+    >
+      <thead>
+        <tr>
+          <th>Name</th>
+          <th>Sex</th>
+          <th>Born</th>
+          <th>Died</th>
+          <th>Mother</th>
+          <th>Father</th>
```
roman-mirzoian, line 21: We can cheat a bit here and improve the code if we create an array of column names and return them in a similar way, I'm currently doing with the `people` array and the `PersonInfo` component below.

roman-mirzoian (changes requested):
The work is done well and only a few points need to be refined.

dffUqp (approved):
GJ✔️

PR #826, opened 2023-07-10
solution

snuber-w:
- [DEMO LINK](https://snuber-w.github.io/react_people-table-basics/)


src/components/PersonLink.tsx
```diff
@@ -0,0 +1,109 @@
...
+      <td>
+        {person.fatherName
+          ? (
+            getParentLink({
+              person,
+              parentType: FATHER,
+              name: person.fatherName,
+              peopleList: people,
+            })
+          )
```
ivankhrsh, line 104: Do you need `parentType: FATHER,` ? You can find person who's name is same as person.fatherName

Also not sure that's it's good practice to calculate it inside markup. You can move it to constant and then pun const here.

src/components/PersonLink.tsx
```diff
@@ -0,0 +1,109 @@
...
+interface Props {
+  person: Person,
+  people: Person[],
+}
+
+export const PersonLink: React.FC<Props> = ({ person, people }) => {
+  const { slug } = useParams();
+  const isFemale = person.sex === FEMALE;
+
+  const { sex, born, died } = person;
```
ivankhrsh, line 59: 👍

src/components/PersonLink.tsx
```diff
@@ -0,0 +1,109 @@
...
+        {isMother
+          ? person.motherName
+          : person.fatherName}
+      </Link>
+    );
+  }
+
+  return `${name}`;
+};
+
```
ivankhrsh, line 49: I think it's better to me it to helpers file

src/App.tsx
```diff
@@ -12,161 +19,53 @@ export const App = () => (
...
-          <div className="box table-container">
-            <Loader />
-
-            <p data-cy="peopleLoadingError" className="has-text-danger">
-              Something went wrong
-            </p>
-
-            <p data-cy="noPeopleMessage">
-              There are no people on the server
-            </p>
```
ivankhrsh, line 45: You forget to add this message if loading is completed, but array is empty

src/App.tsx
```diff
@@ -12,161 +19,53 @@ export const App = () => (
...
+          <Route path="/" element={<h1 className="title">Home page</h1>} />
+          <Route path="/home" element={<Navigate to="/" replace />} />
+          <Route
+            path="/people"
+            element={<PeoplePage />}
+          />
+          <Route
+            path="/people/:slug"
+            element={<PeoplePage />}
+          />
```
ivankhrsh, line 62: You can move this route inside /people so path will be just /:slug

src/components/PersonLink.tsx
```diff
@@ -0,0 +1,109 @@
...
+  const parent = peopleList
+    .find(personToFind => personToFind.name === name);
+
+  if (parent) {
+    return (
+      <Link
+        className={classNames(
+          { 'has-text-danger': isMother },
+        )}
+        to={`/people/${parent.slug}`}
```
ivankhrsh, line 38: Do you need to add /people/? There is better solution to have slug there. It's was in theory videos

ivankhrsh (changes requested):

PR #834, opened 2023-07-10
solution

okosohor:
[DEMO LINK](https://okosohor.github.io/react_people-table-basics/)

### DIFF
--- a/src/components/PeopleList.tsx
+++ b/src/components/PeopleList.tsx
@@ -1,0 +1,97 @@
+import { useEffect, useState } from 'react';
+import { Loader } from './Loader';
+import { getPeople } from '../api';
+import { Person } from '../types';
+import { PersonRow } from './PersonRow';
+
+export const PeopleList = () => {
+  const [people, setPeople] = useState<Person[]>([]);
+  const [loadingError, setLoadingError] = useState(false);
+  const [isLoaded, setIsLoaded] = useState(false);
+
+  const getPreparedPeople = (loadedPeople: Person[]) => {
+    return loadedPeople.map(person => {
+      return {
+        ...person,
+        mother: people
+          .find(person1 => person1.name === person.motherName),
+        father: people
+          .find(person1 => person1.name === person.fatherName),
+      };
+    });
+  };
+
+  useEffect(() => {
+    const fetchData = async () => {
+      try {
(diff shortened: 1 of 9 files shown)


src/components/PeopleList.tsx
```diff
@@ -0,0 +1,97 @@
...
+              className="table is-striped is-hoverable is-narrow is-fullwidth"
+            >
+              <thead>
+                <tr>
+                  <th>Name</th>
+                  <th>Sex</th>
+                  <th>Born</th>
+                  <th>Died</th>
+                  <th>Mother</th>
+                  <th>Father</th>
```
roman-mirzoian, line 78: We can cheat a bit here and put the column names in a separate array outside the component and return it as prepared tags, similar to what we currently do with the `preparedPeople` array and the `PersonRow` component.

src/components/PeopleList.tsx
```diff
@@ -0,0 +1,97 @@
...
+  };
+
+  useEffect(() => {
+    const fetchData = async () => {
+      try {
+        const loadedPeople = await getPeople();
+
+        setPeople(loadedPeople);
+      } catch (error) {
+        setLoadingError(true);
```
roman-mirzoian, line 31: When catching an `error`, it is better to have an understanding of exactly what went wrong and output it to the user or to the console: 
```
setLoadingError(`Something went wrong: ${error.message}`)
```

src/components/PeopleList.tsx
```diff
@@ -0,0 +1,97 @@
...
+import { getPeople } from '../api';
+import { Person } from '../types';
+import { PersonRow } from './PersonRow';
+
+export const PeopleList = () => {
+  const [people, setPeople] = useState<Person[]>([]);
+  const [loadingError, setLoadingError] = useState(false);
+  const [isLoaded, setIsLoaded] = useState(false);
+
+  const getPreparedPeople = (loadedPeople: Person[]) => {
```
roman-mirzoian, line 12: You've already created the `getPreparedPeople` function in `src/utils.ts`, so you should use that function from there.

roman-mirzoian (changes requested):
The task is done well, but there are several points that need to be corrected.

naziamoff (approved):
👍

PR #837, opened 2023-07-11
Solution

julshvets:
[DEMO LINK](https://julshvets.github.io/react_people-table-basics/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,173 +1,32 @@
-import { Loader } from './components/Loader';
-
+import { Routes, Route, Navigate } from 'react-router-dom';
 import './App.scss';
+import { HomePage } from './pages/HomePage';
+import { PeoplePage } from './pages/PeoplePage';
+import { NotFoundPage } from './pages/NotFoundPage';
+import { Navbar } from './components/Navbar';
 
-export const App = () => (
-  <div data-cy="app">
-    <nav
-      data-cy="nav"
-      className="navbar is-fixed-top has-shadow"
-      role="navigation"
-      aria-label="main navigation"
-    >
-      <div className="container">
-        <div className="navbar-brand">
-          <a className="navbar-item" href="#/">
-            Home
-          </a>
+export const App: React.FC = () => {
+  return (
+    <div data-cy="app">
+      <Navbar />
 
-          <a
-            className="navbar-item has-background-grey-lighter"
-            href="#/people"
-          >
-            People
-          </a>
+      <main className="section">
+        <div className="container">
+          <Routes>
+            <Route path="/" element={<HomePage />} />
+            <Route
+              path="/home"
+              element={<Navigate to="/" replace />}
+            />
+            <Route path="/people">
+              <Route index element={<PeoplePage />} />
+              <Route path=":slectedPersonSlug" element={<PeoplePage />} />
+            </Route>
+
+            <Route path="*" element={<NotFoundPage />} />
+          </Routes>
         </div>
-      </div>
-    </nav>
-
-    <main className="section">
-      <div className="container">
-        <h1 className="title">Home Page</h1>
-        <h1 className="title">People Page</h1>
-        <h1 className="title">Page not found</h1>
-
-        <div className="block">
-          <div className="box table-container">
-            <Loader />
-
-            <p data-cy="peopleLoadingError" className="has-text-danger">
-              Something went wrong
-            </p>
-
-            <p data-cy="noPeopleMessage">
-              There are no people on the server
-            </p>
-
-            <table
-              data-cy="peopleTable"
-              className="table is-striped is-hoverable is-narrow is-fullwidth"
-            >
-              <thead>
-                <tr>
-                  <th>Name</th>
-                  <th>Sex</th>
-                  <th>Born</th>
-                  <th>Died</th>
-                  <th>Mother</th>
-                  <th>Father</th>
-                </tr>
-              </thead>
-
-              <tbody>
-                <tr data-cy="person">
-                  <td>
-                    <a href="#/people/jan-van-brussel-1714">
-                      Jan van Brussel
-                    </a>
-                  </td>
-
-                  <td>m</td>
-                  <td>1714</td>
-                  <td>1748</td>
-                  <td>Joanna van Rooten</td>
-                  <td>Jacobus van Brussel</td>
-                </tr>
-
-                <tr data-cy="person">
-                  <td>
-                    <a href="#/people/philibert-haverbeke-1907">
-                      Philibert Haverbeke
-                    </a>
-                  </td>
-
-                  <td>m</td>
-                  <td>1907</td>
-                  <td>1997</td>
-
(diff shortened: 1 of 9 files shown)


src/App.tsx
```diff
@@ -1,173 +1,32 @@
...
+        <div className="container">
+          <Routes>
+            <Route path="/" element={<HomePage />} />
+            <Route
+              path="/home"
+              element={<Navigate to="/" replace />}
+            />
+            <Route path="/people">
+              <Route index element={<PeopleTabPage />} />
+              <Route path=":slectedPersonSlug" element={<PeopleTabPage />} />
```
roman-mirzoian, line 23: There's a typo in the parameter name. It should be `:selectedPersonSlug` instead of `:slectedPersonSlug`.

src/pages/PeopleTabPage.tsx
```diff
@@ -0,0 +1,54 @@
+import { useState, useEffect } from 'react';
+import { Loader } from '../components/Loader';
+import { PeopleTable } from '../components/PeopleTable';
+import { getPeople } from '../api';
+import { Person } from '../types';
+
+export const PeopleTabPage: React.FC = () => {
+  const [peopleFromServer, setPeopleFromServer] = useState<Person[]>([]);
+  const [isLoadin, setIsLoading] = useState(false);
+  const [errorMessage, setErrorMessage] = useState(false);
```
roman-mirzoian, line 10: The variable `errorMessage` is initialized with a boolean value but used to store an error message. It can lead to confusion. Consider renaming it to `hasError` or `errorOccurred`, because now it seems that the error text should be stored there.

src/pages/PeopleTabPage.tsx
```diff
@@ -0,0 +1,54 @@
+import { useState, useEffect } from 'react';
+import { Loader } from '../components/Loader';
+import { PeopleTable } from '../components/PeopleTable';
+import { getPeople } from '../api';
+import { Person } from '../types';
+
+export const PeopleTabPage: React.FC = () => {
+  const [peopleFromServer, setPeopleFromServer] = useState<Person[]>([]);
+  const [isLoadin, setIsLoading] = useState(false);
```
roman-mirzoian, line 9: Be careful when naming, there is a typo in the variable name `isLoadin`. It should be `isLoading`.

src/components/PersonInfo.tsx
```diff
@@ -0,0 +1,70 @@
...
+    slug,
+  } = person;
+
+  const { slectedPersonSlug = '' } = useParams();
+
+  const isSelected = (personSlug: string) => personSlug === slectedPersonSlug;
+
+  return (
+    <tr
+      key={key}
```
roman-mirzoian, line 35: `Key` prop is not required for a `tr` element, unless the row is part of an array being iterated over. In this case, the `key` should be passed at the time of iteration, not inside the individual component.

src/components/PersonInfo.tsx
```diff
@@ -0,0 +1,70 @@
+import { useParams } from 'react-router-dom';
+import classNames from 'classnames';
+import { Person } from '../types';
+import { PersonLink } from './PersonLink';
+
+interface Props {
+  key: string;
```
roman-mirzoian, line 7: Props should not include `key` as a prop type. It's a special prop used by React for lists and is not passed to the components.

src/components/PersonInfo.tsx
```diff
@@ -0,0 +1,70 @@
...
+    name,
+    sex,
+    born,
+    died,
+    fatherName = '-',
+    motherName = '-',
+    slug,
+  } = person;
+
+  const { slectedPersonSlug = '' } = useParams();
```
roman-mirzoian, line 29: Be careful when naming, there is a typo in variable name `slectedPersonSlug`. It should be `selectedPersonSlug`.

src/components/PeopleTable.tsx
```diff
@@ -0,0 +1,42 @@
+import { Person } from '../types';
+import { PersonInfo } from './PersonInfo';
+
+interface Props {
+  people: Person[];
+}
+
+export const PeopleTable: React.FC<Props> = ({ people }) => {
+  const getPersonByName = (name: string | null): Person | null => (name
```
roman-mirzoian, line 9: To improve performance, the `getPersonByName` function can be brought out as a separate helper outside of the component.

src/pages/PeopleTabPage.tsx
```diff
@@ -0,0 +1,54 @@
...
+  const [errorMessage, setErrorMessage] = useState(false);
+
+  const isEmpty = peopleFromServer.length === 0 && !errorMessage;
+
+  useEffect(() => {
+    setIsLoading(true);
+    getPeople()
+      .then(setPeopleFromServer)
+      .catch(() => {
```
roman-mirzoian, line 18: In the `catch` block, you should also catch a possible `error` and return it to the user or output it to the console:
```
.catch((error) => setErrorMessage(`Something went wrong: ${error.message}`))
```
Then it will not be necessary to rename the name of the error state.

src/components/PersonLink.tsx
```diff
@@ -0,0 +1,31 @@
...
+        to={`/people/${slug}`}
+        className={classNames({
+          'has-text-danger': sex === 'f',
+        })}
+      >
+        {name}
+      </Link>
+    ) : (
+      <span>
+        {name || '-'}
```
roman-mirzoian, line 27: It would be cool not to use magic values and put such strings into constants and reuse them:
```
const FEMALE = 'f';
const NOT_SET_VALUE = '-';
```
This will increase the readability of the code and save us from possible errors.
This applies to development in general.

src/components/PeopleTable.tsx
```diff
@@ -0,0 +1,42 @@
...
+      className="table is-striped is-hoverable is-narrow is-fullwidth"
+    >
+      <thead>
+        <tr>
+          <th>Name</th>
+          <th>Sex</th>
+          <th>Born</th>
+          <th>Died</th>
+          <th>Mother</th>
+          <th>Father</th>
```
roman-mirzoian, line 26: We can cheat a little and put the column names in an array outside the component and `.map` on it in the same way as we do with the `people` array.

roman-mirzoian (changes requested):
The application works correctly and the code is written well, but there are several comments that need to be corrected to improve the performance of the code. 
Also, be careful when naming functions and avoid spelling mistakes.

src/components/Navbar.tsx
```diff
@@ -0,0 +1,24 @@
...
+      <div className="navbar-brand">
+        <NavItem
+          title="Home"
+          link="/"
+        />
+
+        <NavItem
+          title="People"
+          link="/people"
+        />
```
dinaminahi, line 20:
```suggestion
        <NavItem title="Home" link="/" />
        <NavItem title="People"  link="/people" />
```

dinaminahi (approved):
Well done 👍

PR #839, opened 2023-07-12
people-table-basics

vladimirshub:
[DEMO LINK](https://vladimirshub.github.io/react_people-table-basics/)

### DIFF
--- a/src/components/Navigation.tsx
+++ b/src/components/Navigation.tsx
@@ -1,0 +1,21 @@
+import React from 'react';
+import { NavigationLink } from './NavigationLink';
+
+export const Navigation: React.FC = () => {
+  return (
+    <nav
+      className="navbar is-fixed-top has-shadow"
+      data-cy="nav"
+      role="navigation"
+      aria-label="main navigation"
+    >
+      <div className="container">
+        <div className="navbar-brand">
+          <NavigationLink to="/" title="Home" />
+          <NavigationLink to="/people" title="People" />
+
+        </div>
+      </div>
+    </nav>
+  );
+};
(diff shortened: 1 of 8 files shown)


src/components/TablePeople/TablePeople.tsx
```diff
@@ -0,0 +1,66 @@
...
+import { Person } from '../../types/Person';
+import { PersonLink } from '../PersonLink';
+
+type Props = {
+  people: Person[],
+  slug: string,
+};
+
+export const TablePeople: React.FC<Props> = ({ people, slug }) => {
+  const isSelected = (person: Person) => person.slug === slug;
```
VitaliyBodnar, line 12: make sense to use useMemo here, because it's essential to performance

src/components/TablePeople/TablePeople.tsx
```diff
@@ -0,0 +1,66 @@
...
+          <th>Father</th>
+        </tr>
+      </thead>
+
+      <tbody>
+        {people.map(person => {
+          const mother = people
+            .find(parent => parent.name === person.motherName);
+          const father = people
+            .find(parent => parent.name === person.fatherName);
```
VitaliyBodnar, line 35: this functionality will be executed every time when the state is changed, so it's bad for the performance
so, make sense to move it to the variables with useMemo

VitaliyBodnar (changes requested):
Well done 🔥
Let's improve your solution 🧑‍💻

src/components/Navigation.tsx
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+import { NavigationLink } from './NavigationLink';
+
+export const Navigation: React.FC = () => {
+  return (
```
etojeDenys, line 5:
```suggestion
export const Navigation: React.FC = () => (
```

etojeDenys (approved):
lgtm

PR #842, opened 2023-07-12
solution

dimoshen:
[DEMO LINK](https://dimoshen.github.io/react_people-table-basics/)

### DIFF
--- a/src/components/PeopleTableRow/PeopleTableRow.tsx
+++ b/src/components/PeopleTableRow/PeopleTableRow.tsx
@@ -1,0 +1,60 @@
+import React from 'react';
+import { Link, useParams } from 'react-router-dom';
+import classNames from 'classnames';
+import { Person } from '../../types';
+
+interface Props {
+  person: Person
+}
+
+export const PeopleTableRow: React.FC<Props> = ({ person }) => {
+  const { slug } = useParams();
+
+  return (
+    <tr
+      data-cy="person"
+      className={classNames({
+        'has-background-warning': person.slug === slug,
+      })}
+    >
+      <td>
+        <Link
+          to={person.slug}
+          className={classNames(
+            { 'has-text-danger': person.sex === 'f' },
+          )}
+        >
+          {person.name}
+        </Link>
+      </td>
+      <td>{person.sex}</td>
+      <td>{person.born}</td>
+      <td>{person.died}</td>
+      <td>
+        {person.mother
+          ? (
+            <Link
+              className="has-text-danger"
+              to={person.mother.slug}
+            >
+              {person.motherName || '-'}
+            </Link>
+          ) : (
+            person.motherName || '-'
+          )}
+      </td>
+      <td>
+        {person.father
+          ? (
+            <Link
+              to={person.father.slug}
+            >
+              {person.fatherName || '-'}
+            </Link>
+          ) : (
+            person.fatherName || '-'
+          )}
+      </td>
+    </tr>
+  );
+};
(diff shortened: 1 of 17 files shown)


src/index.tsx
```diff
@@ -1,14 +1,49 @@
...
+  },
+  {
+    path: '/',
+    element: <App />,
+    errorElement: <NotFoundRedirect />,
+    children: [
+      {
+        path: 'some/not/existing/page',
+        element: <NotFound />,
+      },
```
ulositska, line 27: It's cool that you decided to do a redirect for invalid urls! But if I understand your code correctly, it's supposed to open NotFoundComponent, but currently it's not loading
image


I am not sure how this path works 
path: 'some/not/existing/page',

src/components/MainNav/MainNav.tsx
```diff
@@ -0,0 +1,35 @@
...
+        </NavLink>
+        <NavLink
+          to="/people"
+          className={({ isActive }) => classNames(
+            'navbar-item',
+            { 'has-background-grey-lighter': isActive },
+          )}
+        >
+          People
+        </NavLink>
```
ulositska, line 31: maybe we can make a separate component for that, since they're pretty much identical:thinking:

src/components/PeopleTableRow/PeopleTableRow.tsx
```diff
@@ -0,0 +1,52 @@
...
+    >
+      <td>
+        <Link
+          to={person.slug}
+          className={classNames(
+            { 'has-text-danger': person.sex === 'f' },
+          )}
+        >
+          {person.name}
+        </Link>
```
ulositska, line 29: You already have a TableRowLink, I think, with very slight modifications, It can be used here as well?

src/pages/PeoplePage/PeoplePage.tsx
```diff
@@ -0,0 +1,62 @@
...
+  useEffect(() => {
+    const fetchData = async () => {
+      try {
+        setIsLoading(true);
+        const response = await getPeople();
+
+        const peopleFromServer = peopleWithParentsInfo(response);
+
+        setPeople(peopleFromServer as Person[]);
+        setIsLoading(false);
```
ulositska, line 22: no biggie, but since it's not loading( setIsLoading(false) ) anymore both in the case of success and error, **finally** statement seems fitting, don't you think?

ulositska (approved):
Good solution:fire:
Great implementation of the routing, useParams() and redirect

src/index.tsx
```diff
@@ -1,14 +1,49 @@
...
-  );
+const router = createHashRouter([
+  {
+    path: '/home',
+    element: <Navigate />,
+  },
+  {
+    path: '/',
+    element: <App />,
+    errorElement: <NotFoundRedirect />,
```
dffUqp, line 22: are you sure this is the correct way to use [errorElement](https://reactrouter.com/en/main/route/error-element#errorelement) here

src/components/TableRowLink/TableRowLink.tsx
```diff
@@ -0,0 +1,19 @@
...
+import { Person } from '../../types';
+
+interface Props {
+  parent: Person,
+}
+
+export const TableRowLink: FC<Props> = ({ parent }) => {
+  return (
+    <Link
+      className={classNames({ 'has-text-danger': parent.sex === 'f' })}
```
dffUqp, line 13: let's create variable outside component for "f" because now it's magic value
```jsx
const FEMALE = 'f';
```

dffUqp (changes requested):
Good work but needs a few fixes to make it better:smiley:

PR #847, opened 2023-07-13
Solved

olehmahera:
[DEMO LINK](https://olehmahera.github.io/react_people-table-basics/)


src/components/PeopleTable/PersonLink/PersonLink.tsx
```diff
@@ -0,0 +1,22 @@
...
+};
+
+export const PersonLink: FC<Props> = ({ person }) => {
+  const { sex, slug, name } = person;
+
+  return (
+
+    <NavLink
+      to={`/people/${slug}`}
+      className={cn({ 'has-text-danger': sex === 'f' })}
```
dffUqp, line 17: would be really cool to create a variable outside the component for "f" because now that's the magic value

```jsx
const FEMALE = 'f';
```

src/components/PeopleTable/PeopleTable.tsx
```diff
@@ -0,0 +1,72 @@
...
+              </td>
+              <td>
+                {father
+                  ? <PersonLink person={father} />
+                  : fatherName || '-'}
+              </td>
+            </tr>
+
+          );
+        })}
```
dffUqp, line 68: I think you needed to create a separate component for table row

src/components/PeopleTable/PeopleTable.tsx
```diff
@@ -0,0 +1,72 @@
...
+  const { personSlug = '' } = useParams();
+
+  return (
+    <table
+      data-cy="peopleTable"
+      className="table is-striped is-hoverable is-narrow is-fullwidth"
+    >
+      <thead>
+        <tr>
+          {Object.entries(PeoplePersonHeader).map(([key, value]) => (
```
dffUqp, line 22: we won't reuse `PeoplePersonHeader` , so it's better to just create array with the headers instead of enum

dffUqp (approved):
Looks good:smiley: but always can be better, try not to make these mistakes in the following tasks.

PR #856, opened 2023-07-16
99% solution

Gosyav:
https://vo7kov.github.io/react_people-table-basics/
I can't understand, why the mother links aren't red. I've checked other pulls, I have same code, but it's not working

### DIFF
--- a/src/components/PeopleTable/PeopleTable.tsx
+++ b/src/components/PeopleTable/PeopleTable.tsx
@@ -1,0 +1,37 @@
+import React from 'react';
+
+import { Person } from '../../types';
+import { PersonLink } from '../PersonLink/PersonLink';
+
+type Props = {
+  people: Person[],
+};
+
+export const PeopleTable: React.FC<Props> = ({ people }) => {
+  return (
+    <table
+      data-cy="peopleTable"
+      className="table is-striped is-hoverable is-narrow is-fullwidth"
+    >
+      <thead>
+        <tr>
+          <th>Name</th>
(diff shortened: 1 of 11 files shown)


src/components/PeopleTable/PeopleTable.tsx
```diff
@@ -0,0 +1,37 @@
+import React from 'react';
+
+import { Person } from '../../types';
+import { PersonLink } from '../PersonLink/PersonLink';
+
+type Props = {
+  people: Person[],
+};
+
+export const PeopleTable: React.FC<Props> = ({ people }) => {
```
etojeDenys, line 10: you should map people and find their mothers and fathers

etojeDenys (changes requested):

src/components/PeopleTable/PeopleTable.tsx
Vo7kov: done, thank you!

src/components/PersonLink/PersonLink.tsx
```diff
@@ -0,0 +1,97 @@
...
+                'has-text-danger': sex === 'f',
+              },
+            )}
+          >
+            {name}
+          </Link>
+        )}
+      </td>
+
+      <td>
```
OleziO, line 61: Combine it into one line. Check other similar cases.

OleziO (changes requested):

mykhalenych (approved):

PR #858, opened 2023-07-19
react-people-tab1

fenderukr:
https://fenderukr.github.io/react_people-table-basics/


src/PeoplePage.tsx
```diff
@@ -0,0 +1,81 @@
+import React, { useState } from 'react';
+import { EnumPersonType } from './types/EnumPersonType';
+import { PersonLink } from './PersonLink';
+import { Person } from './types';
+
+interface PeoplePageProps {
+  usersArr: Person[];
+}
```
lerastarynets, line 8: Consider renaming 'usersArr' to a more descriptive name like 'people' or 'persons'. This would make the code more self-explanatory.

src/PeoplePage.tsx
```diff
@@ -0,0 +1,81 @@
...
+
+export const PeoplePage:React.FC<PeoplePageProps> = ({ usersArr }) => {
+  const [selectedPerson, setSelectedPerson] = useState<string | null>(null);
+
+  const handlePersonClick = (personName: string | null) => {
+    setSelectedPerson(personName);
+  };
+
+  const isPresentInTable = (name: string | null) => {
+    return usersArr.some((person) => person.name === name);
```
lerastarynets, line 18: The isPresentInTable function is reiterating over the usersArr on every call. This could lead to performance issues when the array is large. Instead, consider using a lookup object to map person names to their existence in the array.

src/PeoplePage.tsx
```diff
@@ -0,0 +1,81 @@
...
+                person={person}
+                handlePersonClick={handlePersonClick}
+                personType={EnumPersonType.name}
+              />
+            </td>
+
+            <td>{person.sex}</td>
+            <td>{person.born}</td>
+            <td>{person.died}</td>
+            {/* <td>{person.motherName}</td> */}
```
lerastarynets, line 58: The code for the mother's name field is commented out, which means it won't display the mother's name as required in the task description. Uncomment this line or update it to match the pattern used for the other fields.

src/PeoplePage.tsx
```diff
@@ -0,0 +1,81 @@
...
+            <td>
+              <PersonLink
+                person={person}
+                personType={EnumPersonType.motherName}
+                isPresentInTable={isPresentInTable(person.motherName)}
+                handlePersonClick={handlePersonClick}
+              />
+            </td>
+            {/* <td>{person.fatherName}</td> */}
```
lerastarynets, line 67: The code for the father's name field is commented out, which means it won't display the father's name as required in the task description. Uncomment this line or update it to match the pattern used for the other fields.

src/App.tsx
```diff
@@ -1,173 +1,65 @@
...
+        const response = await fetch(
+          'https://mate-academy.github.io/react_people-table/api/people.json',
+        );
+
+        const responseData = await response.json();
+
+        setUsersArr(responseData);
+      } catch (error) {
+        // eslint-disable-next-line no-console
+        console.log('error');
```
lerastarynets, line 28: In production code, avoid console.log statements. To handle errors, consider implementing a user-friendly error message.

src/PersonLink.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+  return (
+    <>
+      {!isPresentInTable
+        ? person[personType] || '-'
+        : (
+          <Link
+            to={`#/people/${person.slug}`}
+            onClick={handleClick}
+            className={classNames('', {
```
lerastarynets, line 31: The 'classNames' function is used with an empty string as its first argument, which doesn't make sense. If you want to conditionally apply the 'has-text-danger' class, you can just pass the object as the only argument to 'classNames'.

src/types/EnumPersonType.tsx
```diff
@@ -0,0 +1,5 @@
+export enum EnumPersonType {
```
lerastarynets, line 1: pls refactor your enum according to [enum naming convention](https://michaelcurrin.github.io/dev-cheatsheets/cheatsheets/typescript/data-types/enums.html#:~:text=Naming%20convention,enum%20tend%20to%20be%20titlecase)

lerastarynets (changes requested):
Nice start! Let's fix and improve your work a little bit)

Also pls don't show empty table on HomePage and PageNotfound
Screenshot at Jul 20 12-21-56

And don't as extra `#/people` to url when a person is selected
Screenshot at Jul 20 12-22-18

src/PeoplePage.tsx
fenderukr: @lerastarynets I am not sure I understand what lookup obj is and how I write it and what it improves here

src/App.tsx
```diff
@@ -1,173 +1,73 @@
...
+        const response = await fetch(
+          'https://mate-academy.github.io/react_people-table/api/people.json',
+        );
+
+        const responseData = await response.json();
+
+        setPeople(responseData);
+      } catch (error) {
+        // eslint-disable-next-line no-alert
+        alert('error');
```
lerastarynets, line 29: Pls use useState to store an error and show it somewhere, don't use alerts/console.logs for such cases

lerastarynets (changes requested):
Well done with fixing comments) Only last thing need to be improved before approval

src/App.tsx
fenderukr: 424267880fd4e2eecbbd66c8c9a37fbb6d7063cc

maxim2310 (approved):
Good work!!!

PR #867, opened 2023-07-24
done

VladyslavKucher:
https://VladyslavKucher.github.io/react_people-table-basics/


src/pages/People.tsx
```diff
@@ -0,0 +1,155 @@
...
+                    </td>
+
+                    <td>{person.sex}</td>
+                    <td>{person.born}</td>
+                    <td>{person.died}</td>
+
+                    {findMother(person) ? (
+                      <td>
+                        <Link
+                          to={`${findMother(person)?.slug}`}
```
lerastarynets, line 124: You are calling the findMother function twice - once for the condition and once for the Link 'to' prop. This is inefficient. Instead, call it once and store the result in a variable to be used in both places.

src/pages/People.tsx
```diff
@@ -0,0 +1,155 @@
...
+                  <th>Born</th>
+                  <th>Died</th>
+                  <th>Mother</th>
+                  <th>Father</th>
+                </tr>
+              </thead>
+
+              <tbody>
+
+                {error === ERROR.EMPTYARRAY && (
```
lerastarynets, line 92: The error message should be placed outside of the table body (tbody) as per HTML semantics. Error messages are not part of table data.

src/pages/People.tsx
```diff
@@ -0,0 +1,155 @@
...
+                        </Link>
+                      </td>
+                    ) : (
+                      <td>{person.motherName || '-'}</td>
+                    )}
+
+                    {findFather(person) ? (
+                      <td>
+                        <Link
+                          to={`${findFather(person)?.slug}`}
```
lerastarynets, line 137: You are calling the findFather function twice - once for the condition and once for the Link 'to' prop. This is inefficient. Instead, call it once and store the result in a variable to be used in both places.

src/pages/People.tsx
```diff
@@ -0,0 +1,155 @@
...
+
+  const findMother = (person: Person) => {
+    if (person.motherName) {
+      return people.find((p) => p.name === person.motherName);
+    }
+
+    return undefined;
+  };
+
+  const findFather = (person: Person) => {
```
lerastarynets, line 32: This func is similar to `findMother` function, it's better to create reusable `findParents`

src/pages/People.tsx
```diff
@@ -0,0 +1,155 @@
...
+
+              <tbody>
+
+                {error === ERROR.EMPTYARRAY && (
+                  <p data-cy="noPeopleMessage">
+                    There are no people on the server
+                  </p>
+                )}
+
+                {people.map(person => (
```
lerastarynets, line 98: pls destucture `person`

lerastarynets (changes requested):
I see you did a hard work) Let's improve it a little bit

src/App.tsx
```diff
@@ -1,173 +1,28 @@
...
-              </tbody>
-            </table>
-          </div>
-        </div>
-      </div>
-    </main>
-  </div>
+  <Router>
+    <Routes>
+      <Route path="/" element={<Page />}>
```
Anastasiia-Svintsova, line 18: The `Page` component is unnecessary and confusing because it is not clear what its purpose is. There should be a `HomePage` instead.

src/pages/ErrorPage.tsx
```diff
@@ -0,0 +1,5 @@
+export const ErrorPage = () => {
```
Anastasiia-Svintsova, line 1: Doesn't match the task rewuirement
image
```suggestion
export const NotFoundPage = () => {
```

Anastasiia-Svintsova (changes requested):
Great job! But some requirements are not met

- This requirement is not implemented. I can't see the PeopleTable component 😢 
image
image

BTW you should create a separate folder for components

lerastarynets (approved):
You've got this, keep pushing forward!

PR #868, opened 2023-07-24
Solution

Ukrainiane-panda:
[DEMO LINK](https://Ukrainiane-panda.github.io/react_people-table-basics/)


src/page/PeoplePage/PeoplePage.tsx
```diff
@@ -0,0 +1,133 @@
...
+import { useParams, NavLink } from 'react-router-dom';
+import cn from 'classnames';
+import { Loader } from '../../components/Loader';
+import { Person } from '../../types';
+import { getPeople } from '../../api';
+
+export const PeoplePage = () => {
+  const { slug } = useParams();
+  const [people, setPeople] = useState<Person[]>([]);
+  const [isLoading, setIsLoading] = useState<boolean>(false);
```
TarasHoliuk, line 11:
```suggestion
  const [isLoading, setIsLoading] = useState(false);
```
just FYI - if you set the default value the type of variable will be defined automatically

src/page/PeoplePage/PeoplePage.tsx
```diff
@@ -0,0 +1,133 @@
...
+  const [isEmpty, setIsEmpty] = useState<boolean>(false);
+  const [isDataFetched, setIsDataFetched] = useState<boolean>(false);
```
TarasHoliuk, line 13: You use this state variable to show the error message. But it's better to add `try catch` block to catch an error and set the proper value to smth like `isError` state variable.
Because now `isDataFetched` looks pretty similar to `isLoading`

src/page/PeoplePage/PeoplePage.tsx
```diff
@@ -0,0 +1,133 @@
...
+import cn from 'classnames';
+import { Loader } from '../../components/Loader';
+import { Person } from '../../types';
+import { getPeople } from '../../api';
+
+export const PeoplePage = () => {
+  const { slug } = useParams();
+  const [people, setPeople] = useState<Person[]>([]);
+  const [isLoading, setIsLoading] = useState<boolean>(false);
+  const [isEmpty, setIsEmpty] = useState<boolean>(false);
```
TarasHoliuk, line 12: you don't need a state for this
just create a common variable to store this - `!people.length`

general rule: no need to create a state variable if you can get its value from other state variables or props

src/page/PeoplePage/PeoplePage.tsx
```diff
@@ -0,0 +1,133 @@
...
+        <div className="box table-container">
+          {isLoading && <Loader />}
+
+          {!isLoading && !isDataFetched
+        && (
+          <p data-cy="peopleLoadingError" className="has-text-danger">
+            Something went wrong
+          </p>
+        )}
+          {isEmpty && isDataFetched
```
TarasHoliuk, line 55: this logic works in the wrong way

src/page/PeoplePage/PeoplePage.tsx
```diff
@@ -0,0 +1,133 @@
...
+    setPeople(arrUsers);
+    setIsDataFetched(true);
+    setIsLoading(false);
+  };
+
+  useEffect(() => {
+    fetchData();
+  }, []);
+
+  const preparedPeopleWithParents = people.map((person) => ({
```
TarasHoliuk, line 35: :+1:

TarasHoliuk (changes requested):
The demo link doesn't work
Please, fix it. Feel free to ask in the chat if you need help

etojeDenys (approved):
lgtm

PR #879, opened 2023-07-30
Solution

RomaSheva1987:
[DEMO LINK](https://RomaSheva1987.github.io/react_people-table-basics/)

### DIFF
--- a/src/components/NavbarLink.tsx
+++ b/src/components/NavbarLink.tsx
@@ -1,0 +1,19 @@
+import { NavLink } from 'react-router-dom';
+import { Pages } from '../types/Pages';
+import { getLinkClass, getLinkStyle } from './utils/Utils';
+
+type Props = {
+  page: string;
+};
+
+export const NavbarLink: React.FC<Props> = ({ page }) => {
+  return (
+    <NavLink
+      to={page === Pages.Home ? '/' : `/${page.toLowerCase()}`}
+      className={getLinkClass}
+      style={getLinkStyle}
+    >
+      {page}
+    </NavLink>
+  );
+};
--- a/src/components/PersonLink.tsx
+++ b/src/components/PersonLink.tsx
@@ -1,0 +1,19 @@
+import React from 'react';
+import classNames from 'classnames';
+import { Link } from 'react-router-dom';
+import { Person } from '../types';
+
+type Props = {
+  person: Person;
+};
+
+export const PersonLink: React.FC<Props> = ({ person }) => (
+  <Link
+    to={`/people/${person.slug}`}
+    className={classNames({
+      'has-text-danger': person.sex === 'f',
+    })}
+  >
+    {person.name}
+  </Link>
+);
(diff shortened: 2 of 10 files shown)


VitaliyBodnar (changes requested):
- the tests are not passed
- linter errors
Screenshot 2023-07-30 at 15 51 25

src/components/PersonLink.tsx
```diff
@@ -0,0 +1,19 @@
...
+
+type Props = {
+  person: Person;
+};
+
+export const PersonLink: React.FC<Props> = ({ person }) => (
+  <Link
+    to={`/people/${person.slug}`}
+    className={classNames({
+      'has-text-danger': person.sex === 'f',
```
VitaliyBodnar, line 14: make sense to move 'f' and 'm' to the enum
please check other places

src/components/peopleTable/PeopleTable.tsx
```diff
@@ -0,0 +1,126 @@
+import { useParams } from 'react-router-dom';
+import classNames from 'classnames';
+import './PeopleTable.scss';
```
VitaliyBodnar, line 3: according to the style guide, import styles should be at the bottom of the imports list

src/components/peopleTable/PeopleTable.tsx
```diff
@@ -0,0 +1,126 @@
...
+                    <th>Died</th>
+                    <th>Mother</th>
+                    <th>Father</th>
+                  </tr>
+                </thead>
+
+                <tbody>
+                  {people.map(person => {
+                    const mother = findParent(person, 'f');
+                    const father = findParent(person, 'm');
```
VitaliyBodnar, line 71: please don't call a function from the template because it will be executed every time when the state change, so make sense to move it to the variable

src/types/Pages.tsx
```diff
@@ -0,0 +1,4 @@
+export enum Pages {
+  Home = 'Home',
+  PEOPLE = 'People',
```
VitaliyBodnar, line 3: enum's elemetns should be in camelCase 
```suggestion
  People = 'People',
```

src/components/NavbarLink.tsx
```diff
@@ -0,0 +1,19 @@
...
+import { getLinkClass, getLinkStyle } from './utils/Utils';
+
+type Props = {
+  page: string;
+};
+
+export const NavbarLink: React.FC<Props> = ({ page }) => {
+  return (
+    <NavLink
+      to={page === Pages.Home ? '/' : `/${page.toLowerCase()}`}
```
VitaliyBodnar, line 12: this condition only applies to two links and does not scale
makes sense to create separate links to "Home" and "People"

VitaliyBodnar (changes requested):
Almost done!

TarasHoliuk (approved):
LGTM :+1:

PR #882, opened 2023-07-31
Solution

adv555:
**[DEMO LINK](https://adv555.github.io/react_people-table-basics)**


irynabolbat (approved):
good job 👍

src/components/PersonCard.tsx
```diff
@@ -0,0 +1,55 @@
...
+type Props = {
+  person: Person;
+  people: Person[];
+};
+
+export const PersonCard: React.FC<Props> = ({ person, people }) => {
+  const { slug } = useParams();
+
+  const mother = people.find((p) => p.name === person.motherName);
+  const father = people.find((p) => p.name === person.fatherName);
```
And678, line 16: Don't do this on every render. Do it on load or in `useMemo`

And678 (changes requested):
Good, but needs to be optimized

src/components/PersonCard.tsx
adv555: Thx 😊 Done ✅

And678 (approved):
:+1:

PR #898, opened 2023-08-04
Solution

outofs:
https://outofs.github.io/react_people-table-basics/
I don't know what a problem with tests, but locally everything is fine:
image


maxyma-sanx (approved):
Good work

src/components/PeopleTable/PeopleTable.tsx
```diff
@@ -0,0 +1,73 @@
...
+                <PersonLink person={person} />
+              </td>
+
+              <td>{person.sex}</td>
+              <td>{person.born}</td>
+              <td>{person.died}</td>
+              <td>
+                {
+                  person.motherName
+                    ? isLinkedPerson(person.motherName, people)
```
And678, line 56: You are doing this `isLinkedPerson` search twice per every person **per every render**.
Memoize it to prevent extra searching. (useMemo)

And678 (changes requested):
Very good, but can be improved

src/components/PeopleRow/PeopleRow.tsx
```diff
@@ -0,0 +1,60 @@
...
+import { PersonLink } from '../PersonLink';
+import { Person } from '../../types';
+
+type Props = {
+  person: Person;
+  slug: string | undefined;
+  people: Person[];
+};
+
+const isLinkedPerson = (name: string, table: Person[]) => {
```
And678, line 13:
```suggestion
const findLinkedPerson = (name: string, table: Person[]) => {
```

And678: This is a func, not a boolean

And678 (approved):
GJ, but check my comment

PR #900, opened 2023-08-04
add solution

IraShtynda:
[DEMO LINK](https://irashtynda.github.io/react_people-table-basics/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,172 +1,25 @@
-import { Loader } from './components/Loader';
-
+import { Routes, Route, Navigate } from 'react-router-dom';
 import './App.scss';
+import { Navbar } from './components/Navbar';
+import { HomePage } from './pages/HomePage';
+import { PeoplePage } from './pages/PeoplePage';
+import { NotFoundPage } from './pages/NotFoundPage';
 
 export const App = () => (
   <div data-cy="app">
-    <nav
-      data-cy="nav"
-      className="navbar is-fixed-top has-shadow"
-      role="navigation"
-      aria-label="main navigation"
-    >
-      <div className="container">
-        <div className="navbar-brand">
-          <a className="navbar-item" href="#/">
-            Home
-          </a>
-
-          <a
-            className="navbar-item has-background-grey-lighter"
-            href="#/people"
-          >
-            People
-          </a>
-        </div>
-      </div>
-    </nav>
+    <Navbar />
 
     <main className="section">
       <div className="container">
-        <h1 className="title">Home Page</h1>
-        <h1 className="title">People Page</h1>
-        <h1 className="title">Page not found</h1>
-
-        <div className="block">
-          <div className="box table-container">
-            <Loader />
-
-            <p data-cy="peopleLoadingError" className="has-text-danger">
-              Something went wrong
-            </p>
-
-            <p data-cy="noPeopleMessage">
-              There are no people on the server
-            </p>
-
-            <table
-              data-cy="peopleTable"
-              className="table is-striped is-hoverable is-narrow is-fullwidth"
-            >
-              <thead>
-                <tr>
-                  <th>Name</th>
-                  <th>Sex</th>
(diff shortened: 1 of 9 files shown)


src/pages/PeoplePage.tsx
```diff
@@ -0,0 +1,67 @@
...
+    const mother = people.find(({ name }) => name === person.motherName);
+    const father = people.find(({ name }) => name === person.fatherName);
+
+    return ({
+      ...person,
+      mother,
+      father,
+    });
+  });
+};
```
NataliGru, line 18: - You can move `peopleWithParents` function to a separate file as a utility (or helper) function.
- You don't need to pass an individual `person` as a parameter, as you need to find `mother` and `father` for all people in the list. Just go through the array of people and find their mothers and fathers.

NataliGru (approved):
Great job✨✨✨

src/App.tsx
```diff
@@ -1,172 +1,25 @@
-import { Loader } from './components/Loader';
-
+import { Routes, Route, Navigate } from 'react-router-dom';
 import './App.scss';
+import { Navbar } from './components/Navbar';
+import { HomePage } from './pages/HomePage';
+import { PeoplePage } from './pages/PeoplePage';
+import { NotFoundPage } from './pages/NotFoundPage';
```
OlEzhA0, line 6: create `index.ts` file in the `pages` folder, export each of your component, and use it here like
```
import { HomePage, PeoplePage, NotFoundPage } from './pages'
```

src/components/Navbar.tsx
```diff
@@ -0,0 +1,25 @@
+import React from 'react';
+import cn from 'classnames';
+import { NavLink } from 'react-router-dom';
+
+const getLinkClass = ({ isActive }: { isActive: boolean }) => cn(
```
OlEzhA0, line 5: move this function to utils

src/components/PeopleTable.tsx
```diff
@@ -0,0 +1,33 @@
+import React from 'react';
+import { Person } from '../types/Person';
+import { PersonRow } from './PersonRow';
+
+type Props = {
+  people: Person[]
+};
+
+export const PeopleTable: React.FC<Props> = ({ people }) => {
+  return (
```
OlEzhA0, line 10: you can miss `return` when returning some value at once

src/components/PersonRow.tsx
```diff
@@ -0,0 +1,46 @@
...
+      <td>{person.born}</td>
+      <td>{person.died}</td>
+
+      <td>
+        {person.mother ? (
+          <PersonLink person={person.mother} />
+        ) : (
+          person.motherName || '-'
+        )}
+      </td>
```
OlEzhA0, line 35: create component for this case and reuse it for father as well

src/pages/PeoplePage.tsx
OlEzhA0: agree with Natali, move this fn to utils

Didn’t catch second comment about passing an individual person as a parameter.

src/pages/PeoplePage.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+export const PeoplePage: React.FC = () => {
+  const [people, setPeople] = useState<Person[]>([]);
+  const [loading, setLoading] = useState(false);
+  const [error, setError] = useState(false);
+
+  useEffect(() => {
+    setLoading(true);
```
OlEzhA0, line 26: set `loading` `true` as initial value to avoid micro jumping of your content
```suggestion
  const [loading, setLoading] = useState(true);
  const [error, setError] = useState(false);

  useEffect(() => {
```

And678 (approved):
Approved, but read @NataliGru's comments

OlEzhA0 (changes requested):
good job! left a few minor comments

anstsot (approved):
Good work, approving:sparkles:

PR #904, opened 2023-08-05
Develop

Valentyna08:
[DEMO LINK](https://Valentyna08.github.io/react_people-table-basics/)

### DIFF
--- a/src/components/Loader/HomePage.tsx
+++ b/src/components/Loader/HomePage.tsx
@@ -1,0 +1,5 @@
+export const HomePage: React.FC = () => {
+  return (
+    <h1 className="title">Home Page</h1>
+  );
+};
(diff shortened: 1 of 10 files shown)


src/components/Loader/HomePage.tsx
And678, line 1: Not clear why this is in the Loader folder.

src/components/PeoplePage.tsx
```diff
@@ -0,0 +1,115 @@
...
+                  <th>Sex</th>
+                  <th>Born</th>
+                  <th>Died</th>
+                  <th>Mother</th>
+                  <th>Father</th>
+                </tr>
+              </thead>
+
+              <tbody>
+                {peopleWithParents(people).map((person) => (
```
And678, line 79: This is executing `peopleWithParents` on every render, put in `useMemo`

And678 (changes requested):
Very nice, only a couple of improvements and this is done :100:

And678 (approved):
:fire:

PR #941, opened 2023-08-13
solution

vetal-hovenko:
[DEMO LINK](https://vetal-hovenko.github.io/react_people-table-basics/)


src/utils/generateNavLink.tsx
```diff
@@ -0,0 +1,11 @@
+import { NavLink } from 'react-router-dom';
+import { Person } from '../types';
+import { setPersonLinkClassName } from './classNamesSetter';
+
+export const generateNavLink = (person: Person) => {
```
anastasiiavorobiova, line 5: It would be better to use object destructuring for accessing object properties. In this case, you can extract 'slug' and 'name' from the 'person' object, which makes the code cleaner and easier to read.

src/PageRouter.tsx
```diff
@@ -0,0 +1,35 @@
...
+import { HomePage } from './components/HomePage';
+import { PeoplePage } from './components/PeoplePage';
+import { NotFoundPage } from './components/NotFoundPage';
+
+export const PageRouter = () => {
+  const location = useLocation();
+  const navigate = useNavigate();
+
+  React.useEffect(() => {
+    if (location.pathname === '/home') {
```
anastasiiavorobiova, line 19: The useEffect redirects /home to / which might not be desired. If /home should be a valid path, this redirection should be removed. Alternatively, if /home should not be a valid path, then this could be better handled with a ```<Redirect>``` component in the ```<Routes>```.

src/utils/fetchPeopleData.ts
```diff
@@ -0,0 +1,12 @@
+export async function fetchPeopleData() {
+  try {
+    const response = await fetch(
+      'https://mate-academy.github.io/react_people-table/api/people.json',
```
anastasiiavorobiova, line 4: Avoid using hardcoded URLs. It's better to store URLs in a separate config file or at least as a constant at the top of the current file.

src/components/PeoplePage.tsx
```diff
@@ -0,0 +1,116 @@
...
+      })
+      .catch(() => {
+        setError(true);
+      })
+      .finally(() => {
+        setIsLoading(false);
+      });
+  }, []);
+
+  const findPerson = (personName: string | null) => {
```
anastasiiavorobiova, line 31: The 'findPerson' function is being used within the rendering of the component, which could impact performance as it runs on every render. Consider using React.useMemo to memoize the function result and improve performance.

src/components/PeoplePage.tsx
```diff
@@ -0,0 +1,116 @@
...
+          {isError && (
+            <p data-cy="peopleLoadingError" className="has-text-danger">
+              Something went wrong
+            </p>
+          )}
+          {peopleData?.length === 0 && !isError && (
+            <p data-cy="noPeopleMessage">There are no people on the server</p>
+          )}
+
+          {!isError && !isLoading
```
anastasiiavorobiova, line 60: The condition in this line is quite complex and might be hard to understand. Consider using descriptive variables for each state condition to improve readability.

src/components/PeoplePage.tsx
```diff
@@ -0,0 +1,116 @@
...
+import { Loader } from './Loader';
+import { Person } from '../types';
+import { fetchPeopleData } from '../utils/fetchPeopleData';
+import { generateNavLink } from '../utils/generateNavLink';
+
+export const PeoplePage = () => {
+  const [peopleData, setPeopleData] = React.useState<Person[] | null>(null);
+  const [isLoading, setIsLoading] = React.useState(false);
+  const [isError, setError] = React.useState(false);
+
```
anastasiiavorobiova, line 13: The variable 'setError' is not consistent with the other state variable names. Consider renaming it to 'setIsError' to maintain consistency.

src/App.tsx
```diff
@@ -1,173 +1,37 @@
...
+    <div data-cy="app">
+      <nav
+        data-cy="nav"
+        className="navbar is-fixed-top has-shadow"
+        role="navigation"
+        aria-label="main navigation"
+      >
+        <div className="container">
+          <div className="navbar-brand">
+            <NavLink className={setNavLinkClassName} to="/">
```
anastasiiavorobiova, line 16: The function `setNavLinkClassName` is being assigned to `className` but it's not being invoked. If it's a function to determine the class name, it should be invoked with necessary parameters.

src/App.tsx
```diff
@@ -1,173 +1,37 @@
...
+              Home
+            </NavLink>
+
+            <NavLink
```
anastasiiavorobiova, line 20: Same issue as above, the function `setNavLinkClassName` is being assigned to `className` but it's not being invoked. If it's a function to determine the class name, it should be invoked with necessary parameters.

anastasiiavorobiova (changes requested):
Great work! Almost perfect!

volodymyr-soltys97 (approved):
Nice work 🔥

PR #942, opened 2023-08-14
Develop

liza-strykharchuk:
[DEMO LINK](https://liza-strykharchuk.github.io/react_people-table-basics/)

i dkn why my npm test is not working


src/App.tsx
```diff
@@ -1,172 +1,26 @@
...
-                  </td>
-                </tr>
-              </tbody>
-            </table>
-          </div>
-        </div>
+    <NavPage />
+
+    <main>
+      <div className="constainer">
```
anastasiiavorobiova, line 15: There is a typo in the class name 'constainer'. The correct is 'container'.

src/components/TablePeople.tsx
```diff
@@ -0,0 +1,42 @@
+import React from 'react';
+
+import { Person } from '../types';
+import { InfoPerson } from './Loader/InfoPerson';
+
+type Prors = {
```
anastasiiavorobiova, line 6: The type name 'Prors' appears to be a typo. It would be better to use a more descriptive name, like 'Props'.

src/components/Loader/InfoPerson.tsx
```diff
@@ -0,0 +1,87 @@
+import React from 'react';
+
+import { Link, useParams } from 'react-router-dom';
+import classNames from 'classnames';
+
+import { Person } from '../../types/Person';
+
+type Propds = {
```
anastasiiavorobiova, line 8: There is a typo on line 8. 'Propds' should be replaced with 'Props'. This will make the code more readable and understandable.

src/PeoplePage.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+export const PeoplePage = () => {
+  const [people, setPeople] = useState<Person[]>([]);
+  const [newError, setNewError] = useState(false);
+  const [load, setLoad] = useState(true);
+
+  useEffect(() => {
+    getPeople()
+      .then((peopleFromServer) => {
+        setLoad(true);
```
anastasiiavorobiova, line 19: The 'setLoad(true)' inside the '.then' block is not necessary since loading is already set to true before the API call and should only be set to false after the data is received or there's an error.

src/PeoplePage.tsx
```diff
@@ -0,0 +1,53 @@
...
+      <div className="box table-container">
+        {load ? (
+          <Loader />
+        ) : (
+          <>
+            {newError && (
+              <LoadingError />
+            )}
+
+            {people.length === 0 && !load && (
```
anastasiiavorobiova, line 43: The '!load' condition is not required in the ternary operator since it's already guaranteed that 'load' is false in this context.

src/PeoplePage.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+  useEffect(() => {
+    getPeople()
+      .then((peopleFromServer) => {
+        setLoad(true);
+        setPeople(peopleFromServer);
+      })
+      .catch((error) => {
+        setNewError(true);
+        throw error;
```
anastasiiavorobiova, line 24: Throwing an error after setting 'setNewError(true)' is not required as it stops the execution of the current function and will not allow for the '.finally' block to be executed and hence, 'setLoad(false)' would not be called.

src/PeoplePage.tsx
```diff
@@ -0,0 +1,53 @@
...
+import { getPeople } from './api';
+import { Loader } from './components/Loader';
+import { NoPeople } from './NoPeople';
+import { LoadingError } from './components/LoadError';
+import { TablePeople } from './components/TablePeople';
+
+
+export const PeoplePage = () => {
+  const [people, setPeople] = useState<Person[]>([]);
+  const [newError, setNewError] = useState(false);
```
anastasiiavorobiova, line 13: Consider adding a generic type here for code consistency

src/PeoplePage.tsx
```diff
@@ -0,0 +1,53 @@
...
+  const [load, setLoad] = useState(true);
```
anastasiiavorobiova, line 14: Consider adding a generic type here for code consistency

anastasiiavorobiova (changes requested):
Great work! Added comments to improve

src/PeoplePage.tsx
liza-strykharchuk: i don't think that it'll be more useful than this

anastasiiavorobiova (approved):
Good work!

PR #950, opened 2023-08-17
added table routing

JasterHim:
[DEMO LINK](https://JasterHim.github.io/react_people-table-basics/)

### DIFF
--- a/src/components/PeopleList.tsx
+++ b/src/components/PeopleList.tsx
@@ -1,0 +1,58 @@
+import React, { useEffect, useState } from 'react';
+import { Person } from '../types';
+import { PeopleItem } from './PeopleItem';
+
+type Props = {
+  people: Person[],
+};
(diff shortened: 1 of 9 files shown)


src/components/PeopleList.tsx
```diff
@@ -0,0 +1,58 @@
...
+export const PeopleList: React.FC<Props> = ({
+  people,
+}) => {
+  const [isNoPeople, setIsNoPeople] = useState(false);
+
+  useEffect(() => {
+    if (!people.length) {
+      setIsNoPeople(true);
+    }
+  }, [people]);
```
etojeDenys, line 18: we don't need state for this logic

let's just check the people length in some condition

etojeDenys (changes requested):

agniya-i (approved):

PR #955, opened 2023-08-19
add task solution

kbekher:
[DEMO LINK](https://kbekher.github.io/react_people-table-basics/)

### DIFF
--- a/src/components/HomePage.tsx
+++ b/src/components/HomePage.tsx
@@ -1,0 +1,5 @@
+export const HomePage = () => {
+  return (
+    <h1 className="title">Home Page</h1>
+  );
+};
--- a/src/components/PersonLink.tsx
+++ b/src/components/PersonLink.tsx
@@ -15,7 +15,7 @@ export const PersonLink: React.FC<Props> = ({ person, people }) => {
     <tr
       data-cy="person"
       className={classNames({
-        'has-background-warning': person.slug === slug,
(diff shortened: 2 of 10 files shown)


src/components/HomePage.tsx
```diff
@@ -0,0 +1,5 @@
+export const HomePage = () => {
+  return (
+    <h1 className="title">Home Page</h1>
+  );
+};
```
VitaliyBodnar, line 5: can be less
```suggestion
export const HomePage = () => ( <h1 className="title">Home Page</h1> );
```

src/components/PersonLink.tsx
```diff
@@ -0,0 +1,60 @@
...
+        </Link>
+      </td>
+
+      <td>{person.sex}</td>
+      <td>{person.born}</td>
+      <td>{person.died}</td>
+      <td>
+        {people.find(p => p.name === person.motherName) ? (
+          <Link
+            to={`/people/${people.find(p => p.name === person.motherName)?.slug}`}
```
VitaliyBodnar, line 38: please don't call a function from the template because it will be executed every time when the state change, so it make sense to move it to the variable
check other places

VitaliyBodnar (changes requested):
Almost done!

VitaliyBondarenko1982 (approved):

PR #961, opened 2023-08-24
add task solution

Luk2asz:
[DEMO LINK](https://Luk2asz.github.io/react_people-table-basics/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,173 +1,70 @@
-import { Loader } from './components/Loader';
+/* eslint-disable no-nested-ternary */
+import {
+  HashRouter, Route, NavLink, Switch, Redirect,
+} from 'react-router-dom';
 
 import './App.scss';
+import { People } from './components/People';
 
-export const App = () => (
-  <div data-cy="app">
-    <nav
-      data-cy="nav"
-      className="navbar is-fixed-top has-shadow"
-      role="navigation"
-      aria-label="main navigation"
-    >
-      <div className="container">
-        <div className="navbar-brand">
-          <a className="navbar-item" href="#/">
-            Home
-          </a>
+export const App = () => {
+  return (
+    <HashRouter>
+      <div data-cy="app">
+        <nav
+          data-cy="nav"
+          className="navbar is-fixed-top has-shadow"
+          role="navigation"
+          aria-label="main navigation"
+        >
+          <div className="container">
+            <div className="navbar-brand">
 
-          <a
-            className="navbar-item has-background-grey-lighter"
-            href="#/people"
-          >
-            People
-          </a>
-        </div>
-      </div>
-    </nav>
+              <NavLink
+                exact
+                to="/"
+                className="navbar-item"
+                activeClassName="navbar-item has-background-grey-lighter"
+              >
+                Home
+              </NavLink>
 
-    <main className="section">
-      <div className="container">
-        <h1 className="title">Home Page</h1>
-        <h1 className="title">People Page</h1>
-        <h1 className="title">Page not found</h1>
+              <NavLink
+                to="/people"
+                className="navbar-item"
+                activeClassName="navbar-item has-background-grey-lighter"
+              >
+                People
+              </NavLink>
 
-        <div className="block">
-          <div className="box table-container">
-            <Loader />
-
-            <p data-cy="peopleLoadingError" className="has-text-danger">
-              Something went wrong
-            </p>
-
-            <p data-cy="noPeopleMessage">
-              There are no people on the server
-            </p>
-
-            <table
-              data-cy="peopleTable"
-              className="table is-striped is-hoverable is-narrow is-fullwidth"
-            >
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -1,173 +1,70 @@
...
+              </NavLink>
+
+            </div>
+          </div>
+        </nav>
+
+        <main className="section">
+          <div className="container">
+            <Switch>
+              <Route exact path="/">
```
darokrk, line 46: I would rather suggest that you to split this render on routes with separate /pages folder structure to be e.g. :

`/pages/HomePage.tsx, /pages/NotFoundPage.tsx, /pages/PeoplePage.tsx`

and pass them on to proper routes

you have described this in task requirements:

"Create the HomePage available at / with just a title Home Page
Create the PeoplePage available at /people with a title People Page
Implement NotFoundPage with a title Page not found that is shown for all the other URLs;"

src/App.tsx
```diff
@@ -1,173 +1,70 @@
...
-                </tr>
-              </tbody>
-            </table>
+import { People } from './components/People';
+
+export const App = () => {
+  return (
+    <HashRouter>
+      <div data-cy="app">
+        <nav
```
darokrk, line 13: You can move that whole logic from `<nav>` tag to `</nav>` into separate component `/components/Navbar.tsx`

src/components/People/People.tsx
```diff
@@ -0,0 +1,164 @@
...
+
+export const People = () => {
+  const { pathname } = useLocation<{ pathname: string }>();
+  const [people, setPeople] = useState<Person[]>([]);
+  const [errorMessage, setErrorMessage] = useState('');
+  const [slugUser, setSlugUser] = useState(pathname.slice(8));
+  const [isLoading, setIsLoading] = useState(false);
+
+  useEffect(() => {
+    const fetchPeople = async () => {
```
darokrk, line 16: We already have this function to get people so please try to use it, dont need to declared new one, `getPeople()` look` import { getPeople } from '../api';`

src/components/People/People.tsx
```diff
@@ -0,0 +1,164 @@
...
+            >
+              Something went wrong
+            </p>
+            <p data-cy="noPeopleMessage">
+              There are no people on the server
+            </p>
+          </>
+        )}
+
+        <table
```
darokrk, line 75: From the task requirements you should create separate component, e.g. `/components/PeopleTable.tsx` and pass props data there.

"Pass people to PeopleTable component and render as a table (see the given markup);"

it can be something like this:

```
<PeopleTable
        people={people}
        errorMessage={errorMessage}
        isLoading={isLoading}
      />
```

src/components/People/People.tsx
```diff
@@ -0,0 +1,164 @@
...
+              <th>Sex</th>
+              <th>Born</th>
+              <th>Died</th>
+              <th>Mother</th>
+              <th>Father</th>
+            </tr>
+          </thead>
+
+          <tbody>
+            {people.map(onePeople => (
```
darokrk, line 91: this also can be splited into separate component like: `/components/PersonItem.tsx` and used directly in new `PeopleTable.tsx` component. It will reduce the logic a bit, e.g

```
 {people.map((person) => (
                <PersonItem key={person.slug} person={person} />
              ))}
```

darokrk (changes requested):
UI looks good 👍🏽  but code need to be refactored a bit 👨🏽‍💻  please make sure that all requirements from task description are done

src/components/People/index.tsx
```diff
@@ -0,0 +1 @@
+export * from './People';
```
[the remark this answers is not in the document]
Luk2asz: But it is exactly like Loader.tsx and index.tsx in loader component. Loader.tsx and index.tsx was at start, I do not change them.

src/components/People/People.tsx
Luk2asz: I use getPeople() in code below and just give some new things(setIsLoading, catch error)
useEffect(() => {
    const fetchPeople = async () => {
      setIsLoading(true);
      try {
        const data = await getPeople();

        setPeople(data);
      } catch (error) {
        setErrorMessage('Unable to load people');
      }

      setIsLoading(false);
    };

    fetchPeople();
  }, []);

Radoslaw-Czerniawski (changes requested):
According to the requirements you should:
- use [Navigate](https://reactrouter.com/en/6.15.0/components/navigate) component
- create the `<PersonLink person={person} />` and use it for all existing people;

Luk2asz:
> According to the requirements you should:
> 
> * use [Navigate](https://reactrouter.com/en/6.15.0/components/navigate) component
> * create the `<PersonLink person={person} />` and use it for all existing people;

When I try to use Navigate, I have error like below: 
image
I have react-router-dom@5.3.4 and I used Redirect instead Navigate

PR #967, opened 2023-08-30
add task solution

akozlovska:
[DEMO LINK](https://akozlovska.github.io/react_people-table-basics/)

### DIFF
--- a/src/components/PeopleTable/PeopleTable.tsx
+++ b/src/components/PeopleTable/PeopleTable.tsx
@@ -1,0 +1,102 @@
+import cn from 'classnames';
+import { useEffect } from 'react';
+import { useParams } from 'react-router-dom';
+import { Person } from '../../types';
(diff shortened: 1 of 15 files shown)


src/components/PeopleTable/PeopleTable.tsx
```diff
@@ -0,0 +1,102 @@
...
+
+            <td>{person.sex}</td>
+            <td>{person.born}</td>
+            <td>{person.died}</td>
+            <td>
+              {person.motherName ? (
+                <>
+                  {getPersonByName(person.motherName) ? (
+                    <Personlink
+                      person={getPersonByName(person.motherName) as Person}
```
etojeDenys, line 68: let's move it to some var when you mapping the array to avoid redundant function call

etojeDenys (changes requested):

TarasHoliuk (approved):

PR #970, opened 2023-09-02
Solution

s3rg4sh:
https://s3rg4sh.github.io/react_people-table-basics/

### DIFF
--- a/src/components/PeoplePage/PeoplePage.tsx
+++ b/src/components/PeoplePage/PeoplePage.tsx
@@ -1,0 +1,54 @@
+import { useEffect, useState } from 'react';
+import { getPeople } from '../../api';
+import { Person } from '../../types';
+import { Loader } from '../Loader';
+import { PeopleTable } from '../PeopleTable';
+
+export const PeoplePage: React.FC = () => {
+  const [people, setPeople] = useState<Person[]>([]);
+  const [isLoading, setIsLoading] = useState(true);
+  const [isError, setIsError] = useState(false);
+
+  useEffect(() => {
+    getPeople()
+      .then(setPeople)
+      .catch(() => setIsError(true))
+      .finally(() => setIsLoading(false));
+  }, []);
+
+  return (
+    <>
+      <h1 className="title">People Page</h1>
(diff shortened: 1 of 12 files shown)


src/components/PeopleTable/PeopleTable.tsx
```diff
@@ -0,0 +1,95 @@
...
+                        <Personlink
+                          person={getPersonByName(person.fatherName) as Person}
+                        />
+                      )
+                      : (
+                        <>
+                          {person.fatherName}
+                        </>
+                      )}
+                  </>
```
VitaliyBondarenko1982, line 84: here and for person mother name same logic.
suggest to make function and use in both places. it will increase readibility

src/components/PeoplePage/PeoplePage.tsx
```diff
@@ -0,0 +1,54 @@
...
+                      ? (
+                        <PeopleTable people={people} />
+                      )
+                      : (
+                        <p data-cy="noPeopleMessage">
+                          There are no people on the server
+                        </p>
+                      )}
+                  </>
+                )}
```
VitaliyBondarenko1982, line 47: a lot of nested ternary. let's refactor and improve readibility of code.
for example write in constant some results of nested ternary and use in jsx.

VitaliyBondarenko1982 (changes requested):
Good job.
Suggest some refactoring.

src/components/PeopleTable/PeopleTable.tsx
```diff
@@ -0,0 +1,83 @@
...
+  };
+
+  const getPersonParent = (name: string | null): JSX.Element => {
+    return (name ? (
+      <>
+        {getPersonByName(name)
+          ? (
+            <Personlink
+              person={getPersonByName(name) as Person}
+            />
```
etojeDenys, line 24: let's save getPersonByName result in some var to avoid redundant function call

etojeDenys (changes requested):

etojeDenys (approved):
good job

PR #971, opened 2023-09-05
solution

vlkzmn:
[DEMO LINK](https://vlkzmn.github.io/react_people-table-basics/)

### DIFF
--- a/src/components/PeopleTable.tsx
+++ b/src/components/PeopleTable.tsx
@@ -1,0 +1,62 @@
+import { useParams } from 'react-router-dom';
+import classNames from 'classnames';
+import { Person } from '../types';
+import { PersonLink } from './PersonLink';
+
+type Props = {
+  people: Person[];
+};
+
+export const PeopleTable: React.FC<Props> = ({ people }) => {
+  const { slug } = useParams();
+
+  function getPerson(name: string | null) {
+    const person = people.find(item => item.name === name);
+
+    if (person) {
+      return <PersonLink person={person} />;
+    }
+
+    return name || '-';
+  }
+
+  return (
+    <table
+      data-cy="peopleTable"
+      className="table is-striped is-hoverable is-narrow is-fullwidth"
+    >
+      <thead>
+        <tr>
+          <th>Name</th>
+          <th>Sex</th>
+          <th>Born</th>
+          <th>Died</th>
+          <th>Mother</th>
+          <th>Father</th>
+        </tr>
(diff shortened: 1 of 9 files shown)


src/pages/PeoplePage.tsx
```diff
@@ -0,0 +1,49 @@
...
+import { PeopleTable } from '../components/PeopleTable';
+
+export const PeoplePage = () => {
+  const [people, setPeople] = useState<Person[]>([]);
+  const [errorMessage, setErrorMessage] = useState('');
+  const [loading, setLoading] = useState(true);
+
+  useEffect(() => {
+    getPeople()
+      .then((response) => {
```
Anastasiia-Svintsova, line 14: The fetched data should be checked before being set to the state. It might not be an array which could cause issues in the PeopleTable component.

src/components/PeopleTable.tsx
```diff
@@ -0,0 +1,62 @@
...
+          <th>Mother</th>
+          <th>Father</th>
+        </tr>
+      </thead>
+
+      <tbody>
+        {people.map(person => (
+          <tr
+            data-cy="person"
+            key={person.slug}
```
Anastasiia-Svintsova, line 43: Don't forget about objects destructuring

src/components/PeopleTable.tsx
```diff
@@ -0,0 +1,62 @@
...
+          >
+            <td>
+              <PersonLink person={person} />
+            </td>
+
+            <td>{person.sex}</td>
+            <td>{person.born}</td>
+            <td>{person.died}</td>
+            <td>{getPerson(person.motherName)}</td>
+            <td>{getPerson(person.fatherName)}</td>
```
Anastasiia-Svintsova, line 56: Do calculations outside the JSX code to keep your JSX code clean and focused on rendering elements

Anastasiia-Svintsova (changes requested):
Almost done!

src/components/PeopleTable.tsx
vlkzmn: Не розумію що ти маєшь на увазі, поясни будь ласка як це зробити.

src/pages/PeoplePage.tsx
vlkzmn: типу ось так?
.then(response => {
      if (!response.ok) {
        throw new Error(`error`)
      }

      return response.json();
    })

PR #974, opened 2023-09-09
people table

victor-buldenko:
[demo](https://victor-buldenko.github.io/react_people-table-basics/)


src/pages/People.tsx
```diff
@@ -0,0 +1,151 @@
...
+export const People: React.FC = () => {
+  const [loading, setLoading] = useState(false);
+  const [error, setError] = useState(false);
+  const { human } = useParams();
+  const { people, setPeople } = peopleStore();
+
+  useEffect(() => {
+    setLoading(true);
+    getPeople()
+      .then((res) => setPeople(res))
```
etojeDenys, line 17:
```suggestion
      .then(setPeople)
```

src/pages/People.tsx
```diff
@@ -0,0 +1,151 @@
...
+                        <td>{died}</td>
+                        <td>{parentLink(motherName)}</td>
+                        <td>{parentLink(fatherName)}</td>
+                      </tr>
+                    );
+                  })}
+                  <tr
+                    data-cy="person"
+                    // className="has-background-warning"
+                    // className="has-text-danger"
```
etojeDenys, line 128: remove all comments

src/pages/People.tsx
```diff
@@ -0,0 +1,151 @@
...
+                        Jan van Brussel
+                      </a>
+                    </td>
+
+                    <td>m</td>
+                    <td>1714</td>
+                    <td>1748</td>
+                    <td>Joanna van Rooten</td>
+                    <td>Jacobus van Brussel</td>
+                  </tr>
```
etojeDenys, line 143: looks like redundant item

etojeDenys (changes requested):

src/Store.ts
```diff
@@ -0,0 +1,13 @@
+import { create } from 'zustand';
```
TarasHoliuk, line 1: :+1: If you add any npm library you should list it in the package.json, in the `dependencies` array

generally, if you run the command like this `npm install zustand` it should happen automatically

src/pages/People.tsx
```diff
@@ -0,0 +1,132 @@
...
+  const { human } = useParams();
+  const { people, setPeople } = peopleStore();
+
+  useEffect(() => {
+    setLoading(true);
+    getPeople()
+      .then(setPeople)
+      .catch(err => {
+        setError(true);
+        Error(err.message);
```
TarasHoliuk, line 20:
```suggestion
```
it does nothing, so you can delete it
you already handle an error (you have `error` state), so no need to `throw new Error` or do something like this

src/pages/People.tsx
```diff
@@ -0,0 +1,132 @@
...
+      <div className="block">
+        <div className="box table-container">
+
+          {error && (
+            <p data-cy="peopleLoadingError" className="has-text-danger">
+              Something went wrong
+            </p>
+          )}
+
+          {people.length === 0
```
TarasHoliuk, line 61:
```suggestion
          {people.length === 0 && !loading
```
this message shouldn't be visible during loading:
image

TarasHoliuk (changes requested):

src/pages/People.tsx
victor-buldenko: It's just a show for the 'do something' error.

src/Store.ts
victor-buldenko: but it has not happened

SanyaBratashchuk (approved):

PR #986, opened 2023-09-24
would love some advise on line 40 of table.tsx and why do i get error…

szymonpachucki:
…s with Route exact    [DEMO LINK](https://<your_account>.github.io/react_people-table-basics/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,173 +1,76 @@
-import { Loader } from './components/Loader';
-
+import React, { useEffect, useState } from 'react';
+import {
+  HashRouter, Route, NavLink,
+} from 'react-router-dom';
 import './App.scss';
+import { Table } from './components/table';
+import { getPeople } from './api';
+import { Person } from './types';
 
-export const App = () => (
-  <div data-cy="app">
-    <nav
-      data-cy="nav"
-      className="navbar is-fixed-top has-shadow"
-      role="navigation"
-      aria-label="main navigation"
-    >
-      <div className="container">
-        <div className="navbar-brand">
-          <a className="navbar-item" href="#/">
-            Home
-          </a>
+export const App: React.FC = () => {
+  const [activePage, setActivePage] = useState('');
+  const [people, setPeople] = useState<Person[]>([]);
 
-          <a
-            className="navbar-item has-background-grey-lighter"
-            href="#/people"
-          >
-            People
-          </a>
-        </div>
-      </div>
-    </nav>
+  useEffect(() => {
+    if (activePage === 'people') {
+      getPeople()
+        .then(data => {
+          return setPeople(data);
+        });
+    }
+  }, [activePage]);
 
-    <main className="section">
-      <div className="container">
-        <h1 className="title">Home Page</h1>
-        <h1 className="title">People Page</h1>
-        <h1 className="title">Page not found</h1>
+  return (
+    <HashRouter>
+      <div data-cy="app">
+        <nav
+          data-cy="nav"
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,173 +1,76 @@
...
+            <Route exact path="/people">
+              <div className="block">
+                <div className="box table-container">
+                  <h1 className="title">People Page</h1>
+                  <Table people={people} />
+                </div>
+              </div>
+            </Route>
+
+            <Route exact path="/">
```
pawel-peksa, line 64: react router v6 doesn't support exact anymore.

// old - v5 `<Route exact path="/" component={Home} />`
// new - v6 `<Route path="/" element={<Home />} />`

As stated in their documentation:

You don't need to use an exact prop on <Route path="/"> anymore. This is because all paths match exactly by default. If you want to match more of the URL because you have child routes use a trailing * as in <Route path="users/*">.

src/components/table.tsx
```diff
@@ -0,0 +1,70 @@
...
+                people.find((person) => person.name === p.mother) ? (
+                  <a href={`#/people/${people.find((person) => person.name === p.mother)?.slug}`}>
+                    {p.mother}
+                  </a>
+                ) : (
+                  p.mother
+                )
+              ) : (
+                '-'
+              )}
```
pawel-peksa, line 50: In general, such nesting of ternary operator is really hard to read and should be avoided. Please also remove this code repetition:
`people.find((person) => person.name === p.mother`
We should not itarate over array more than necessary.

src/App.tsx
```diff
@@ -1,173 +1,76 @@
...
+export const App: React.FC = () => {
+  const [activePage, setActivePage] = useState('');
+  const [people, setPeople] = useState<Person[]>([]);
+
+  useEffect(() => {
+    if (activePage === 'people') {
+      getPeople()
+        .then(data => {
+          return setPeople(data);
+        });
```
pawel-peksa, line 19:
```suggestion
        .then(data => setPeople(data));
```
or even:

```suggestion
        .then(setPeople);
```

src/App.tsx
```diff
@@ -1,173 +1,76 @@
...
+import { getPeople } from './api';
+import { Person } from './types';
+
+export const App: React.FC = () => {
+  const [activePage, setActivePage] = useState('');
+  const [people, setPeople] = useState<Person[]>([]);
+
+  useEffect(() => {
+    if (activePage === 'people') {
+      getPeople()
```
pawel-peksa, line 16: Please add error handling

src/App.tsx
```diff
@@ -1,173 +1,76 @@
...
-                  </td>
-                </tr>
-              </tbody>
-            </table>
+import { Table } from './components/table';
+import { getPeople } from './api';
+import { Person } from './types';
+
+export const App: React.FC = () => {
+  const [activePage, setActivePage] = useState('');
```
pawel-peksa, line 11: Let's narrow all posibilties here to a type union

HaniaNassalska (commented):
I can’t find the page - wrong address. I noticed you are in the middle of changes requested by the mentor, so I’ll only leave a comment 🙂

choeqq (approved):
@szymonpachucki code looks fine, every change requested was implemented, please make sure to provide correct DEMO url, it would be easier for us to check your work

szymonpachucki:
> @szymonpachucki code looks fine, every change requested was implemented, please make sure to provide correct DEMO url, it would be easier for us to check your work

Updated the link, but it does not display anything

PR #1020, opened 2023-10-01
Develop

neovendano:
[DEMO LINK](https://neovendano.github.io/react_people-table-basics/)


serkrops (changes requested):
I think, that you need to split your routing part to separate file and make navbar in main App file. And you need to get your main content block back.

PR #1031, opened 2023-10-02
task react_people-table-basics

vatatan:
[DEMO LINK](https://vatatan.github.io/react_people-table-basics/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,173 +1,47 @@
-import { Loader } from './components/Loader';
-
+import React, { useEffect, useState } from 'react';
 import './App.scss';
 
-export const App = () => (
-  <div data-cy="app">
-    <nav
-      data-cy="nav"
-      className="navbar is-fixed-top has-shadow"
-      role="navigation"
-      aria-label="main navigation"
-    >
-      <div className="container">
-        <div className="navbar-brand">
-          <a className="navbar-item" href="#/">
-            Home
-          </a>
+import { Navigate, Route, Routes } from 'react-router-dom';
+import { getPeople } from './api';
+import { Person } from './types';
+import { HeadOfTable } from './components/TableHead/TableHead';
+import { Navigation } from './components/Navigation/Navigation';
+import { PeoplePage } from './components/PeoplePage/PeoplePage';
 
-          <a
-            className="navbar-item has-background-grey-lighter"
-            href="#/people"
-          >
-            People
-          </a>
+export const App: React.FC = () => {
+  const [people, setPeople] = useState<Person[]>([]);
+
+  useEffect(() => {
+    getPeople()
+      .then(setPeople);
+  }, []);
+
+  return (
+    <div data-cy="app">
+      <HeadOfTable />
+      <main className="section">
+        <div className="container">
+          <Routes>
+            <Route path="/" element={<Navigation />}>
+              <Route index element={<h1 className="title">Home Page</h1>} />
+              <Route
+                path="*"
+                element={<h1 className="title">Page not found</h1>}
+              />
+              <Route
+                path="people"
+                element={<PeoplePage people={people} setPeople={setPeople} />}
+              >
+                <Route
+                  path=":personSlug"
+                  element={<PeoplePage people={people} setPeople={setPeople} />}
+                />
+              </Route>
+              <Route path="home" element={<Navigate to="/" replace />} />
+            </Route>
+          </Routes>
         </div>
-      </div>
-    </nav>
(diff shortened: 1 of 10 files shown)


src/App.tsx
```diff
@@ -1,173 +1,47 @@
...
-                    </a>
-                  </td>
-                </tr>
-              </tbody>
-            </table>
-          </div>
+import { Navigate, Route, Routes } from 'react-router-dom';
+import { getPeople } from './api';
+import { Person } from './types';
+import { HeadOfTable } from './components/TableHead/TableHead';
```
BudnikOleksii, line 7: Use the same name for your component as a file

src/App.tsx
```diff
@@ -1,173 +1,47 @@
...
+import { Navigation } from './components/Navigation/Navigation';
+import { PeoplePage } from './components/PeoplePage/PeoplePage';
+
+export const App: React.FC = () => {
+  const [people, setPeople] = useState<Person[]>([]);
+
+  useEffect(() => {
+    getPeople()
+      .then(setPeople);
```
BudnikOleksii, line 16: Add a catch block to handle errors and notify the user

src/App.tsx
```diff
@@ -1,173 +1,47 @@
...
+  }, []);
+
+  return (
+    <div data-cy="app">
+      <HeadOfTable />
+      <main className="section">
+        <div className="container">
+          <Routes>
+            <Route path="/" element={<Navigation />}>
+              <Route index element={<h1 className="title">Home Page</h1>} />
```
BudnikOleksii, line 26: Create a separate component for your home page, the same for not found page

src/App.tsx
```diff
@@ -1,173 +1,47 @@
...
+                element={<h1 className="title">Page not found</h1>}
+              />
+              <Route
+                path="people"
+                element={<PeoplePage people={people} setPeople={setPeople} />}
+              >
+                <Route
+                  path=":personSlug"
+                  element={<PeoplePage people={people} setPeople={setPeople} />}
+                />
```
BudnikOleksii, line 38: Get and set people it's better to do only on their page, as you actually need them only on a specific page

src/App.tsx
```diff
@@ -1,173 +1,47 @@
...
+  const [people, setPeople] = useState<Person[]>([]);
+
+  useEffect(() => {
+    getPeople()
+      .then(setPeople);
+  }, []);
+
+  return (
+    <div data-cy="app">
+      <HeadOfTable />
```
BudnikOleksii, line 21: You have the `<HeadOfTable />` inside the nav too, I think you don't need to render it here

src/components/Navigation/Navigation.tsx
```diff
@@ -0,0 +1,24 @@
...
+    <>
+      <nav
+        data-cy="nav"
+        className="navbar is-fixed-top has-shadow"
+        role="navigation"
+        aria-label="main navigation"
+      >
+        <div className="container">
+          <div className="navbar-brand">
+            <HeadOfTable />
```
BudnikOleksii, line 16: As I see `HeadOfTable` is the nav too, so you have the same nav inside nav 🤯

src/components/People/PeopleTable.tsx
```diff
@@ -0,0 +1,72 @@
...
+              </td>
+
+              <td>{person.sex}</td>
+              <td>{person.born}</td>
+              <td>{person.died}</td>
+              <td>{getParentsLink(person.motherName)}</td>
+              <td>{getParentsLink(person.fatherName)}</td>
+            </tr>
+          </tbody>
+        ))}
```
BudnikOleksii, line 68:
```suggestion
          <tbody>
            {people.map((person) => (
               <PersonRow key={person.slug} person={person} />
            ))}
          </tbody>
```
You shouldn't create a `tbody` for every person and let's create a separate `Person` or `PersonRow` component.

src/components/PeoplePage/PeoplePage.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+    (async () => {
+      try {
+        setPeople(await getPeople());
+      } catch {
+        setHasError(true);
+      } finally {
+        setLoading(false);
+      }
+    })();
```
BudnikOleksii, line 28: As you get people here, you actually don't need to fetch them inside the App component

BudnikOleksii (changes requested):
Good work 👍 
Let's get rid of unnecessary things and slightly refactor your solution

PR #1033, opened 2023-10-02
Develop

cardinal1991:
[DEMO LINK](https://cardinal1991.github.io/react_people-table-basics/)

### DIFF
--- a/src/components/Navbar.tsx
+++ b/src/components/Navbar.tsx
@@ -1,0 +1,37 @@
+/* eslint-disable no-nested-ternary */
+import { NavLink } from 'react-router-dom';
+
+export const Navbar: React.FC = () => {
+  return (
+    <>
+      <nav
+        className="navbar is-fixed-top has-shadow"
+        role="navigation"
+        aria-label="main navigation"
+        data-cy="nav"
+      >
+        <div className="container">
+          <div className="navbar-brand">
+            <NavLink
+              to="/"
+              className={({ isActive, isPending }) => (
+                isPending ? 'navbar-item' : isActive
+                  ? 'navbar-item has-background-grey-lighter' : 'navbar-item')}
+            >
+              Home
+            </NavLink>
+            <NavLink
+              to="/people"
+              className={({ isActive, isPending }) => (
+                isPending ? 'navbar-item' : isActive
+                  ? 'navbar-item has-background-grey-lighter' : 'navbar-item')}
+            >
+              People
+            </NavLink>
+          </div>
+        </div>
+      </nav>
+
+    </>
+  );
+};
--- a/src/components/People.tsx
+++ b/src/components/People.tsx
@@ -1,0 +1,136 @@
+/* eslint-disable jsx-a11y/anchor-is-valid */
+/* eslint-disable jsx-a11y/no-static-element-interactions */
+/* eslint-disable jsx-a11y/no-noninteractive-element-interactions */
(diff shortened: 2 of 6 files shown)


src/components/People.tsx
```diff
@@ -0,0 +1,164 @@
+/* eslint-disable prefer-template */
+/* eslint-disable jsx-a11y/anchor-is-valid */
+/* eslint-disable jsx-a11y/no-static-element-interactions */
+/* eslint-disable jsx-a11y/no-noninteractive-element-interactions */
+/* eslint-disable jsx-a11y/click-events-have-key-events */
```
choeqq, line 5: try to disable eslint as little as possible, it's actually pretty usefel 😄

src/components/People.tsx
```diff
@@ -0,0 +1,164 @@
...
+export const People: React.FC = () => {
+  const [people, setPeople] = useState<Person[]>();
+  const [isClicked, setIsClicked] = useState<string | null>('');
+  const [isError, setIsError] = useState<string | null>(null);
+
+  getPeople().then((res) => {
+    setPeople(res);
+  }).catch((error) => {
+    setIsError(error);
+  });
```
choeqq, line 21: this should be done in an `useEffect` ❗ ❗ ❗ 

or in some kind of event, but NEVER just fetch items inside your comonent directly

src/components/People.tsx
```diff
@@ -0,0 +1,164 @@
...
+                      >
+                        <td>
+                          {/* <a
+                        href="#/people/jan-van-brussel-1714"
+                        onClick={() => {
+                          setIsClicked(person.name);
+                        }}
+                      >
+                        {person.name}
+                      </a> */}
```
choeqq, line 87: cleanup

choeqq (changes requested):

src/components/Navbar.tsx
```diff
@@ -0,0 +1,37 @@
+/* eslint-disable no-nested-ternary */
+import { NavLink } from 'react-router-dom';
+
+export const Navbar: React.FC = () => {
+  return (
+    <>
```
pawel-peksa, line 6: We don't need Fragment here

src/components/People.tsx
pawel-peksa: +1

src/components/People.tsx
pawel-peksa: +1

src/components/People.tsx
pawel-peksa: +1

src/components/People.tsx
```diff
@@ -0,0 +1,164 @@
...
+                                e.preventDefault();
+                                setIsClicked(person?.motherName);
+                              }}
+                            >
+                              {person.motherName}
+                            </a>
+                          )}
+                          {person.motherName
+                          && !names?.includes(person.motherName) && (
+                            <>
```
pawel-peksa, line 115: Do we really need these Fragments here?

pawel-peksa (changes requested):
Good job! There ale still comments from previous reviewer to be resolved. I've added couple more. You could think about refactoring of conditional rendering in People.tsx. It looks really convoluted

anpiet (approved):
Looks good to me.

choeqq (approved):
LGTM ✅

PR #1041, opened 2023-10-03
Develop

Matylda-Lewandowska:
[DEMO LINK](https://matylda-lewandowska.github.io/react_people-table-basics/)

### DIFF
--- a/src/components/PeopleTable/PeopleTable.tsx
+++ b/src/components/PeopleTable/PeopleTable.tsx
@@ -1,0 +1,58 @@
+import { useEffect, useState } from 'react';
+import { Loader } from '../Loader';
+import { getPeople } from '../../api';
+import { Person } from '../../types';
+import { Table } from '../Table/Table';
+
+export const PeopleTable = () => {
+  const [people, setPeople] = useState<Person[]>([]);
+  const [isError, setIsError] = useState<boolean>(false);
+  const [isLoading, setIsLoading] = useState<boolean>(false);
+
+  useEffect(() => {
+    setIsLoading(true);
+    getPeople()
+      .then((response) => setPeople(response))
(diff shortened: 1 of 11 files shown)


ogospadzipalina (approved):

src/components/PeopleTable/PeopleTable.tsx
```diff
@@ -0,0 +1,58 @@
+import { useEffect, useState } from 'react';
+import { Loader } from '../Loader';
+import { getPeople } from '../../api';
+import { Person } from '../../types';
+import { Table } from '../Table/Table';
+
+export const PeopleTable = () => {
+  const [people, setPeople] = useState<Person[]>([]);
+  const [isError, setIsError] = useState<boolean>(false);
+  const [isLoading, setIsLoading] = useState<boolean>(false);
```
mbruhler, line 10: When you go to this page, you can see the UI glitch. There is unnecessary "There are no people on the server", for few milliseconds and then it hides. The solution is to set loading state to true and then just set it to false when data is loaded.

Then the UI won't glitch

mbruhler: I mean, set it initially to `true`

`const [isLoading, setIsLoading] = useState<boolean>(true);`

src/components/PeopleTable/PeopleTable.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+  useEffect(() => {
+    setIsLoading(true);
```
mbruhler, line 13: then you don't need this

mbruhler (changes requested):
Hi,
Great work on your side :)
Just a minor UI change and it will be approved!

choeqq (approved):
LGTM ✅

PR #1047, opened 2023-10-03
Develop

Dolnys:
[DEMO LINK](https://Dolnys.github.io/react_people-table-basics/)


choeqq (approved):
index files which are exporting each component, don't have to be `.tsx` - `ts` is enough

PR #1049, opened 2023-10-03
solution

PawelP2001:
[DEMO LINK](https://PawelP2001.github.io/react_people-table-basics/)

### DIFF
--- a/src/components/PeopleTable.tsx
+++ b/src/components/PeopleTable.tsx
@@ -1,0 +1,69 @@
+import { useEffect, useState } from 'react';
+import { Loader } from './Loader';
+import { getPeople } from '../api';
+import { Person } from '../types';
+import { PersonItem } from './PersonItem';
+
+export const PeopleTable = () => {
+  const [people, setPeople] = useState<Person[]>([]);
+  const [isError, setIsError] = useState<boolean>(false);
+  const [isLoading, setIsLoading] = useState<boolean>(false);
+
+  useEffect(() => {
+    setIsLoading(true);
+    getPeople()
+      .then((response) => setPeople(response))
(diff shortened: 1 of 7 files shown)


src/components/PeopleTable.tsx
```diff
@@ -0,0 +1,69 @@
+import { useEffect, useState } from 'react';
+import { Loader } from './Loader';
+import { getPeople } from '../api';
+import { Person } from '../types';
+import { PersonItem } from './PersonItem';
+
+export const PeopleTable = () => {
```
choeqq, line 7: you styles are little off - this table shouldn't be 100% of the page's width

choeqq (changes requested):

src/components/PeopleTable.tsx
```diff
@@ -0,0 +1,69 @@
...
+  const [people, setPeople] = useState<Person[]>([]);
+  const [isError, setIsError] = useState<boolean>(false);
+  const [isLoading, setIsLoading] = useState<boolean>(false);
+
+  useEffect(() => {
+    setIsLoading(true);
```
choeqq, line 13: you can set loading to true initially, then remove it `.finally`

choeqq (changes requested):
One last change and you're good to go ✅

src/components/PeopleTable.tsx
PawelP2001: @choeqq isnt it whats already done?

PawelP2001: @choeqq i've changed it so that it initially is true and after resolving gets changed to false

choeqq (approved):
LGTM ✅

PR #1053, opened 2023-10-04
Solution

Nightycake:
[DEMO LINK](https://Nightycake.github.io/react_people-table-basics/)

### DIFF
--- a/src/components/PeopleTable.tsx
+++ b/src/components/PeopleTable.tsx
@@ -1,0 +1,87 @@
+import { useEffect } from 'react';
+import cn from 'classnames';
+import { useParams } from 'react-router-dom';
+import { Person } from '../types';
+import { PeopleLink } from './PeopleLink';
+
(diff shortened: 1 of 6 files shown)


src/components/PeopleTable.tsx
```diff
@@ -0,0 +1,87 @@
...
+
+    if (selectedPerson) {
+      selectedPerson.scrollIntoView({
+        block: 'center',
+        behavior: 'smooth',
+      });
+    } else {
+      window.scrollTo({ top: 0, behavior: 'smooth' });
+    }
+  }, [slug]);
```
choeqq, line 29: since you're mapping the whole table you can get selected element from (person), and you should almost never use `querySelector` in react, then this effect will be unnecessary

choeqq (changes requested):

choeqq (approved):
LGTM ✅

PR #1072, opened 2023-10-16
Solution

NataliiaHen:
[DEMO LINK](https://NataliiaHen.github.io/react_people-table-basics/)

### DIFF
--- a/src/components/PeopleTable/PeopleTable.tsx
+++ b/src/components/PeopleTable/PeopleTable.tsx
@@ -1,0 +1,87 @@
+import React from 'react';
+import classNames from 'classnames';
+import { useParams } from 'react-router-dom';
+import { Person } from '../../types';
+import { PersonLink } from '../PersonLink/PersonLink';
+
+type Props = {
+  people: Person[],
+};
+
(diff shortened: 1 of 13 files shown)


src/components/PeopleTable/PeopleTable.tsx
```diff
@@ -0,0 +1,87 @@
...
+            motherName,
+          } = person;
+
+          const motherContent = motherName
+            ? getParentLink(motherName)
+            : '-';
+
+          const fatherContent = fatherName
+            ? getParentLink(fatherName)
+            : '-';
```
VitaliyBondarenko1982, line 60: we can avoid this two similar ternary and decide whatr render in getParentLink.
just add check on start of getParentLink util
```
if (!name) {
 return '-';
}
```
and call in jsx getParentLink
```
<td>{getParentLink(motherName)}</td>
<td>{getParentLink(fatherName)}</td>
```

VitaliyBondarenko1982 (changes requested):
Looks good.
Please check one minor comment.

VitaliyBondarenko1982 (approved):

PR #1087, opened 2023-10-29
add task solution

ydashko:
[DEMO LINK](https://ydashko.github.io/react_people-table-basics/)

### DIFF
--- a/src/components/PersonLink/PersonLink.tsx
+++ b/src/components/PersonLink/PersonLink.tsx
@@ -1,0 +1,77 @@
+import React from 'react';
+import classNames from 'classnames';
+import { Link, useParams } from 'react-router-dom';
+import { Person } from '../../types';
+
+type Props = {
+  person: Person;
+};
+
+export const PersonLink: React.FC<Props> = ({ person }) => {
+  const {
+    name,
+    sex,
+    born,
+    died,
+    fatherName,
+    motherName,
+    slug,
+    father,
+    mother,
+  } = person;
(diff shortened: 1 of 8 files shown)


src/components/PersonLink/PersonLink.tsx
```diff
@@ -0,0 +1,77 @@
...
+        <Link to={`${slug}`} className={getClassByGender(sex)}>
+          {name}
+        </Link>
+      </td>
+
+      <td>{sex}</td>
+      <td>{born}</td>
+      <td>{died}</td>
+      <td>
+        {getParentsInfo(motherName, mother)}
```
VitaliyBodnar, line 70: please don't call a function from the template because it will be executed every time when the state change, so make sense to move it to the variable

src/components/PersonLink/PersonLink.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+  const getClassByGender = (gender: string) => {
+    return classNames({
+      'has-text-danger': gender === 'f',
+    });
+  };
+
+  const getParentsInfo = (
+    parentName: string | null,
+    personFromData: Person | undefined,
```
VitaliyBodnar, line 33:
```suggestion
    personFromData:? Person,
```

VitaliyBodnar (changes requested):
Almost done!

volodymyr-soltys97 (approved):
Well done!

PR #1097, opened 2023-11-06
solution

OlegPopovych:
Hi,
[DEMO LINK](https://OlegPopovych.github.io/react_people-table-basics/)

### DIFF
--- a/src/components/PeoplePage/PeoplePage.tsx
+++ b/src/components/PeoplePage/PeoplePage.tsx
@@ -57,7 +57,7 @@ export const PeoplePage = () => {
       return <ErrorLoadPage />;
     }
 
-    if (!people.length && !isLoading && isError) {
+    if (people.length === 0 && !isLoading) {
       return (
         <p data-cy="noPeopleMessage">
           There are no people on the server


src/App.tsx
```diff
@@ -1,173 +1,66 @@
...
+            <Route path="people">
+              <Route index element={<PeoplePage />} />
+              <Route path=":slug" element={<PeoplePage />} />
+            </Route>
+
+            <Route
+              path="*"
+              element={<NotFoundPage />}
+            />
+          </Routes>
```
yulabakun, line 61: Would be nice to create Root file for that and use <Outlet /> here, so you will separate routes logic and all other jsx

src/components/ErrorLoadPage/ErrorLoadPage.tsx
```diff
@@ -0,0 +1,7 @@
+export const ErrorLoadPage = () => {
+  return (
+    <p data-cy="peopleLoadingError" className="has-text-danger">
+      Something went wrong
+    </p>
+  );
+};
```
yulabakun, line 7: This doesn't need to be a separate page, it is just error message to show instead of people table.

src/components/PeoplePage/PeoplePage.tsx
```diff
@@ -0,0 +1,148 @@
...
+      return <ErrorLoadPage />;
+    }
+
+    if (!people.length) {
+      return (
+        <p data-cy="noPeopleMessage">
+          There are no people on the server
+        </p>
+      );
+    }
```
yulabakun, line 66: Also maybe you need to check the condition for that error. 
Now when I open people page it shows this error just for a moment, before the loader appears. 
I think happens because when the component is mounted you initialize people with empty array.

image

yulabakun (approved):
Everything works well, my comments are just suggestions

src/components/PeoplePage/PeoplePage.tsx
OlegPopovych: Thanks a lot!

src/components/PeoplePage/PeoplePage.tsx
```diff
@@ -0,0 +1,148 @@
...
+
+    try {
+      const peopleFromServer = await getPeople();
+
+      const personWithParents = peopleFromServer
+        .map(person => {
+          const mother = peopleFromServer
+            .find(parent => parent.name === person.motherName);
+          const father = peopleFromServer
+            .find(parent => parent.name === person.fatherName);
```
polosanya, line 30: Pay attention that find method can return undefined.
Would be great to set the reserve value here

src/components/PeoplePage/PeoplePage.tsx
```diff
@@ -0,0 +1,148 @@
...
+
+            const father = () => {
+              if (!person.fatherName) {
+                return '-';
+              }
+
+              return people
+                .find(per => per.fatherName === person.father?.name)
+                ? <PersoneLink person={person.father} />
+                : person.fatherName;
```
polosanya, line 105: This part can be simplified

polosanya (approved):
Nice! I'm approving it but please pay attention to the comments)

src/components/PeoplePage/PeoplePage.tsx
OlegPopovych: Thanks, it's ok, I used it when I was defining what to render in the table. As for me, there is no reserve value, only existing object or false value

src/components/PeoplePage/PeoplePage.tsx
OlegPopovych: I have no idea how to deal with it. I always have two steps to making a decision about what to render: 
1. does the parent name exist or not
2. specify, whether this parent is in the people list or not.

Could you give me a hint?

PR #1106, opened 2023-11-09
add solution

Helen-arch:
[DEMO LINK](https://Helen-arch.github.io/react_people-table-basics/)


src/App.tsx
```diff
@@ -1,173 +1,15 @@
...
-          </div>
-        </div>
-      </div>
-    </main>
+    <Outlet />
   </div>
 );
+// function classNames(arg0: string, arg1: { 'is-active': boolean; }) {
+//   throw new Error('Function not implemented.');
+// }
```
polosanya, line 15: Remove the comments

src/components/PeopleTable/PeopleTable.tsx
```diff
@@ -0,0 +1,100 @@
...
+            born,
+            died,
+            motherName,
+            fatherName,
+          } = person;
+
+          const mother = people.find(parent => parent.name === motherName);
+          const father = people.find(parent => parent.name === fatherName);
+
+          const mum = mother && people.includes(mother)
```
polosanya, line 43: There is no need to check if people array includes mother as you already found it, so it definitely includes

polosanya (changes requested):
Almost perfect!

src/pages/PeoplePage.tsx
```diff
@@ -0,0 +1,57 @@
+import { useEffect, useState } from 'react';
+import { Loader } from '../components/Loader';
+import { PeopleTable } from '../components/PeopleTable/PeopleTable';
+import { Person } from '../types/Person';
+import { getPeople } from '../api';
+
+export const PeoplePage = () => {
+  const [people, setPeople] = useState<Person[]>([]);
+  const [isError, setIsError] = useState(false);
+  const [loading, setLoading] = useState(false);
```
OlegPopovych, line 10: this component shows the table just for a moment at first render before set `isLoading` to `true`, try to set `true` value as default

OlegPopovych (approved):
Nice!

polosanya (approved):
Well done! 🚀

PR #1116, opened 2023-11-14
Solution

YaroslavYarynych:
Hi! Check how it works:
[DEMO LINK](https://YaroslavYarynych.github.io/react_people-table-basics/)

### DIFF
--- a/src/Root.tsx
+++ b/src/Root.tsx
@@ -1,0 +1,26 @@
+import {
+  Routes,
+  Route,
+  HashRouter as Router,
+  Navigate,
+} from 'react-router-dom';
+import { App } from './App';
+import { HomePage } from './components/Loader/homePage/HomePage';
+import { PeoplePage } from './components/Loader/people/PeoplePage';
+
+export const Root = () => (
+  <Router>
+    <Routes>
+      <Route path="/" element={<App />}>
+        <Route index element={<HomePage />} />
+        <Route path="home" element={<Navigate to="/" replace />} />
+        <Route path="people">
+          <Route index element={<PeoplePage />} />
+          <Route path=":personSlug?" element={<PeoplePage />} />
+        </Route>
+
+        <Route path="*" element={<h1 className="title">Page not found</h1>} />
+      </Route>
+    </Routes>
+  </Router>
+);
--- a/src/components/Loader/people/PeoplePage.tsx
+++ b/src/components/Loader/people/PeoplePage.tsx
@@ -1,0 +1,97 @@
+/* eslint-disable max-len */
+import { useEffect, useState } from 'react';
+import { useParams } from 'react-router-dom';
+import { getPeople } from '../../../api';
+import { Person } from '../../../types';
+import { Loader } from '../Loader';
+import { PeopleRow } from './PeopleRow';
+
+export const PeoplePage = () => {
+  const [people, setPeople] = useState<Person[]>([]);
+  const [isLoading, setIsLoading] = useState(true);
+  const [isError, setIsError] = useState('');
+
+  useEffect(() => {
+    const fetchData = async () => {
+      setIsLoading(true);
+      try {
+        const peopleArray = await getPeople();
+
+        const peopleFromServer = peopleArray.map(person => {
+          const mother = peopleArray.find(p => p.name === person.motherName);
+          const father = peopleArray.find(p => p.name === person.fatherName);
+
+          return {
+            ...person,
+            mother,
+            father,
+          };
+        });
+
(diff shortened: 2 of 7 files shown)


src/Root.tsx
```diff
@@ -0,0 +1,26 @@
...
+    <Routes>
+      <Route path="/" element={<App />}>
+        <Route index element={<HomePage />} />
+        <Route path="home" element={<Navigate to="/" replace />} />
+        <Route path="people">
+          <Route index element={<PeoplePage />} />
+          <Route path=":personSlug?" element={<PeoplePage />} />
+        </Route>
+
+        <Route path="*" element={<h1 className="title">Page not found</h1>} />
```
OlhaKoziuk, line 22: Not nessesary, but on task we should Implement NotFoundPage, as I understand, so just show the title probably is not enough

src/components/Loader/people/PeoplePage.tsx
```diff
@@ -0,0 +1,97 @@
...
+import { PeopleRow } from './PeopleRow';
+
+export const PeoplePage = () => {
+  const [people, setPeople] = useState<Person[]>([]);
+  const [isLoading, setIsLoading] = useState(true);
+  const [isError, setIsError] = useState('');
+
+  useEffect(() => {
+    const fetchData = async () => {
+      setIsLoading(true);
```
OlhaKoziuk, line 16: not nessesary, we have already had  setIsLoading(true) by default

src/components/Loader/people/PeoplePage.tsx
```diff
@@ -0,0 +1,97 @@
...
+      try {
+        const peopleArray = await getPeople();
+
```
OlhaKoziuk, line 19: that is not good idea to use Array in our variables, for example, we can use peopleFromServer

src/components/Loader/people/PeoplePage.tsx
```diff
@@ -0,0 +1,97 @@
...
+                <th>Died</th>
+                <th>Mother</th>
+                <th>Father</th>
+              </tr>
+            </thead>
+
+            <tbody>
+              {people.map(person => (
+                <PeopleRow person={person} key={person.slug} slug={personSlug} />
+              ))}
```
OlhaKoziuk, line 89: we should wrap the line in case we have too much attributes for better readable

src/components/Loader/people/PeopleRow.tsx
```diff
@@ -0,0 +1,49 @@
...
+  return (
+    <tr
+      data-cy="person"
+      key={person.slug}
+      className={cn({
+        'has-background-warning': slug === person.slug,
+      })}
+    >
+      <td>
+        <Link
```
OlhaKoziuk, line 22: As I understand it should be a separate component (check create the <PersonLink person={person} /> and use it for all existing people;)

src/components/Loader/people/PeopleRow.tsx
```diff
@@ -0,0 +1,49 @@
...
+      </td>
+
+      <td>{person.sex}</td>
+      <td>{person.born}</td>
+      <td>{person.died}</td>
+      <td>
+        {person.mother ? (
+          <PersonLink person={person.mother} />
+        )
+          : person.motherName || '-'}
```
OlhaKoziuk, line 39: probably it will be better to wrap the text in a tag for example p, but it`s on you

src/components/Loader/people/PeopleRow.tsx
```diff
@@ -0,0 +1,49 @@
...
+      </td>
+      <td>
+        {person.father ? (
+          <PersonLink person={person.father} />
+        )
+          : person.fatherName || '-'}
```
OlhaKoziuk, line 45: same here

OlhaKoziuk (approved):
Amaizing! Keep going!

src/components/Loader/people/PeoplePage.tsx
```diff
@@ -0,0 +1,100 @@
...
+  }, []);
+
+  const { personSlug } = useParams();
+
+  return (
+    <main className="section">
+      <h1 className="title">People Page</h1>
+
+      <div className="container">
+        {((isLoading || isError) || people.length === 0) && (
```
polosanya, line 48: Would be great to create variables for such complicated conditions

polosanya (changes requested):
Pay attention to the styles.
Page doesn't match the working example(
image
image

StasSokolov (approved):
Looks good to me, i'm approving it🔥

PR #1129, opened 2023-11-15
task solution

YePolishchuk:
[DEMO LINK](https://YePolishchuk.github.io/react_people-table-basics/)

### DIFF
--- a/src/components/PeoplePage/PeoplePage.tsx
+++ b/src/components/PeoplePage/PeoplePage.tsx
@@ -1,0 +1,51 @@
+import { useEffect, useState } from 'react';
+import { Loader } from '../Loader';
+import { Person } from '../../types';
+import { getPeople } from '../../api';
+import { EmptyPeopleList } from '../EmptyPeopleList/EmptyPeopleList';
+import { ErrorMessage } from '../ErrorMessage/ErrorMessage';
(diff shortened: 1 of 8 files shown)


src/components/PeoplePage/PeoplePage.tsx
```diff
@@ -0,0 +1,51 @@
...
+    const mother = people.find(pers => pers.name === person.motherName);
+    const father = people.find(pers => pers.name === person.fatherName);
+
+    return { ...person, mother, father };
+  });
+};
+
+export const PeoplePage = () => {
+  const [people, setPeople] = useState<Person[]>([]);
+  const [loader, setLoader] = useState(false);
```
OlegPopovych, line 20: Try to set `true` as the default value for `loader` to prevent showing `<EmptyPeopleList />` component after the page render.

OlegPopovych (approved):
Greate!

src/components/PeoplePage/PeoplePage.tsx
StasSokolov, line 20: will be more informative use `is/has` prefix for boolean state

StasSokolov (changes requested):
Almost masterpiece, let's improve your code a bit more🔥

StasSokolov (approved):
Great job, i'm approving it🔥

PR #1130, opened 2023-11-15
solution

Ayogoharo:
[DEMO LINK](https://ayogoharo.github.io/react_people-table-basics/)

### DIFF
--- a/src/components/PeoplePage/PeoplePage.tsx
+++ b/src/components/PeoplePage/PeoplePage.tsx
@@ -1,0 +1,57 @@
+import { useEffect, useState } from 'react';
+
+import { Person } from '../../types';
+import { getPeople } from '../../api';
+import { Loader } from '../Loader';
+import { PeopleTablet } from '../PeopleTablet';
+
+export const PeoplePage = () => {
+  const [people, setPeople] = useState<Person[]>([]);
+  const [error, setError] = useState(false);
+  const [isLoading, setIsLoading] = useState(false);
+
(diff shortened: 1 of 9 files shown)


src/components/PeoplePage/PeoplePage.tsx
```diff
@@ -0,0 +1,57 @@
+import { useEffect, useState } from 'react';
+
+import { Person } from '../../types';
+import { getPeople } from '../../api';
+import { Loader } from '../Loader';
+import { PeopleTablet } from '../PeopleTablet';
+
+export const PeoplePage = () => {
+  const [people, setPeople] = useState<Person[]>([]);
+  const [error, setError] = useState(false);
```
polosanya, line 10:
```suggestion
  const [isError, setIsError] = useState(false);
```

src/components/PeopleTablet/PeopleTablet.tsx
```diff
@@ -0,0 +1,77 @@
...
+            sex,
+            born,
+            died,
+            motherName,
+            fatherName,
+            slug,
+          } = person;
+
+          const mother: Person | undefined = people
+            .find(prsn => prsn.name === motherName);
```
polosanya, line 42:
```suggestion
          const mother = people
            .find(prsn => prsn.name === motherName);
```
There is actually no need to specify the type. Also, you could set '-' as the reserve value

polosanya (approved):
Good job!

PR #1135, opened 2023-11-16
add task solution

o-drozzdyk:
[DEMO LINK](https://o-drozzdyk.github.io/react_people-table-basics/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,173 +1,53 @@
-import { Loader } from './components/Loader';
-
 import './App.scss';
+import cn from 'classnames';
+import {
+  NavLink,
+  Navigate,
+  Route,
+  Routes,
+} from 'react-router-dom';
+import { HomePage } from './pages/HomePage';
+import { PeoplePage } from './pages/PeoplePage';
+import { NotFoundPage } from './pages/NotFoundPage';
 
-export const App = () => (
-  <div data-cy="app">
-    <nav
-      data-cy="nav"
-      className="navbar is-fixed-top has-shadow"
-      role="navigation"
-      aria-label="main navigation"
-    >
-      <div className="container">
-        <div className="navbar-brand">
-          <a className="navbar-item" href="#/">
-            Home
-          </a>
+export const App = () => {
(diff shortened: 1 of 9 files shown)


src/App.tsx
```diff
@@ -1,173 +1,53 @@
...
-);
+      </nav>
+
+      <main className="section">
+        <div className="container">
+          <Routes>
+            <Route path="/" element={<HomePage />} />
+            <Route path="/home" element={<Navigate to="/" replace />} />
+            <Route path="/people/:selectedSlug?" element={<PeoplePage />} />
+            <Route path="*" element={<NotFoundPage />} />
```
ivaZaiets, line 47: Better create some new ```Root.tsx``` file and write this logic in that place

src/pages/PeoplePage.tsx
```diff
@@ -0,0 +1,52 @@
...
+            ? <Loader />
+            : (
+              <>
+                {isError && (
+                  <p data-cy="peopleLoadingError" className="has-text-danger">
+                    Something went wrong
+                  </p>
+                )}
+
+                {people.length
```
ivaZaiets, line 39: About errors: our page should look like on the first screenshot, when ```network offline```. In your website I see smth like that (look on the second screenshot). But I see that your tests passed, but I don't know why (hahah), because my tests thought that is wrong variant

image


image

ivaZaiets (approved):
Good job! I approve this because almost all test passed, but please, fix this error with errors. And I see test which didn't pass. I had the same problem and in my case I create new project with identical code and then my test passed. So if you want you also can do smth like that. I see this logic work correctly, but thid problem with test (I think). Good luck!

src/App.tsx
```diff
@@ -1,173 +1,53 @@
...
+            <NavLink to="/people" className={getLinkClass}>
+              People
+            </NavLink>
           </div>
         </div>
-      </div>
-    </main>
-  </div>
-);
+      </nav>
```
superpooperxxx, line 39: Move to separate component

superpooperxxx (approved):
🔥

PR #1162, opened 2023-12-01
solution

Roman-Kulchytskyi:
https://roman-kulchytskyi.github.io/react_people-table-basics/


src/components/HomePage/HomePage.tsx
SanyaBratashchuk, line 1: Create separate folder for pages

SanyaBratashchuk (changes requested):

PR #1171, opened 2023-12-07
solution

xnquaint:
[DEMO LINK](https://xnquaint.github.io/react_people-table-basics/)

### DIFF
--- a/src/components/PeopleTable/PeopleTable.tsx
+++ b/src/components/PeopleTable/PeopleTable.tsx
@@ -1,0 +1,83 @@
+import classNames from 'classnames';
+import { Link, useParams } from 'react-router-dom';
+import { Person } from '../../types';
+
+export const PeopleTable = ({ people }: { people: Person[] }) => {
+  const { personSlug } = useParams();
+
+  return (
+    <table
+      data-cy="peopleTable"
+      className="table is-striped is-hoverable is-narrow is-fullwidth"
+    >
+      <thead>
+        <tr>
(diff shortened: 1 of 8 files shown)


src/components/PeopleTable/PeopleTable.tsx
```diff
@@ -0,0 +1,83 @@
...
+          motherName,
+          slug,
+          mother = people.find(user => user.name === motherName),
+          father = people.find(user => user.name === fatherName),
+        }) => (
+          <tr
+            data-cy="person"
+            className={classNames({
+              'has-background-warning': personSlug === slug,
+            })}
```
etojeDenys, line 40: do not forget about key prop

src/components/PersonLink/PersonLink.tsx
```diff
@@ -0,0 +1,66 @@
...
+      <td>{born}</td>
+      <td>{died}</td>
+      <td>
+        {mother ? (
+          <Link
+            to={`../${mother.slug}`}
+            className="has-text-danger"
+          >
+            {mother.name}
+          </Link>
```
TarasHoliuk, line 48: This part could be the `PersonLink` component (which you reuse for name, mother and father if they exists)

But currently there is an entire row in the component, so it should have been named differently (PersonRow, etc)

because now current structure doesn't match naming

TarasHoliuk (approved):
Approved :heavy_check_mark: 
But do not forget to read one comment FYI (it's important)

PR #1198, opened 2023-12-26
finish task

yefimov77:
[DEMO LINK](https://yefimov77.github.io/react_people-table-basics/)

### DIFF
--- a/src/components/PeopleTable/PeopleTable.tsx
+++ b/src/components/PeopleTable/PeopleTable.tsx
@@ -1,0 +1,41 @@
+/// <reference lib="dom" />
+import { getPeopleWithParents } from '../../helpers';
+import { Person } from '../../types';
+import { PersonRow } from '../Person/PersonRow';
+
+interface Props {
+  people: Person[],
+}
+
+export const PeopleTable: React.FC<Props> = ({ people }) => {
+  const peopleWithParents = getPeopleWithParents(people);
+
+  return (
+    <div className="block">
+      <div className="box table-container">
+
+        <table
+          data-cy="peopleTable"
+          className="table is-striped is-hoverable is-narrow is-fullwidth"
+        >
+          <thead>
+            <tr>
+              <th>Name</th>
+              <th>Sex</th>
+              <th>Born</th>
+              <th>Died</th>
+              <th>Mother</th>
+              <th>Father</th>
+            </tr>
+          </thead>
+
+          <tbody>
+            {peopleWithParents.map(person => (
+              <PersonRow person={person} key={person.slug} />
+            ))}
+          </tbody>
+        </table>
+      </div>
+    </div>
+  );
+};
--- a/src/components/Person/PersonRow.tsx
+++ b/src/components/Person/PersonRow.tsx
@@ -1,0 +1,68 @@
+import cn from 'classnames';
+import { Link, useParams } from 'react-router-dom';
+import { Person } from '../../types';
+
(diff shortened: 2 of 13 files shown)


src/components/PeopleTable/PeopleTable.tsx
```diff
@@ -0,0 +1,41 @@
+/// <reference lib="dom" />
```
roman-mirzoian, line 1: Try to add the necessary types so as not to bypass the typescript

src/components/Person/PersonRow.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+  return (
+    <tr
+      data-cy="person"
+      className={cn({ 'has-background-warning': personSlug === slug })}
+    >
+      <td>
+        <Link
+          to={`/people/${slug}`}
+          className={cn({ 'has-text-danger': sex === 'f' })}
```
roman-mirzoian, line 32: Avoid using "magic values"

src/helpers.ts
```diff
@@ -0,0 +1,18 @@
+import { Person } from './types';
+
+export const getPeopleWithParents = (people: Person[]) => {
+  return people.map(person => {
+    const mother = people.find(mom => person.motherName === mom.name || null);
+    const father = people.find(dad => person.fatherName === dad.name || null);
+    const motherName = !person.motherName ? '-' : person.motherName;
+    const fatherName = !person.fatherName ? '-' : person.fatherName;
```
roman-mirzoian, line 8: To avoid using magic values:
```suggestion
    const NOT_SET_VALUE = '-';
    ...
    const fatherName = !person.fatherName ? NOT_SET_VALUE : person.fatherName;
```

src/helpers.ts
```diff
@@ -0,0 +1,18 @@
+import { Person } from './types';
+
+export const getPeopleWithParents = (people: Person[]) => {
+  return people.map(person => {
+    const mother = people.find(mom => person.motherName === mom.name || null);
+    const father = people.find(dad => person.fatherName === dad.name || null);
```
roman-mirzoian, line 6: You don't need to manually return `null`, the `find` method will return `undefined` if it doesn't find anything:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find

roman-mirzoian (changes requested):
Good job, but there is quite a bit of work to be done.

src/components/Person/PersonRow.tsx
```diff
@@ -19,6 +19,7 @@ export const PersonRow: React.FC<Props> = ({ person }) => {
     father,
   } = person;
 
+  const isFemale = sex === 'f';
```
roman-mirzoian, line 22: It's good that you moved it to a separate variable, but the "magic value" is still used - what is `f`?
It would be good to create an enum in which to describe what `female: 'f', male: 'm'`

roman-mirzoian (approved):
Well done, but pay attention to the comment 👍

src/components/Person/PersonRow.tsx
yefimov77: > It's good that you moved it to a separate variable, but the "magic value" is still used - what is `f`? It would be good to create an enum in which to describe what `female: 'f', male: 'm'`

i`m sorry, i misunderstood correctly, what`s magic value , and thought it`s bout expression in condition... :c

PR #1223, opened 2024-01-05
Solution

zaura333:
[DEMO LINK](https://zaura333.github.io/react_people-table-basics/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,173 +1,29 @@
-import { Loader } from './components/Loader';
-
+import { Navigate, Route, Routes } from 'react-router-dom';
 import './App.scss';
+import { Navigation } from './components/Navigation';
+import { HomePage } from './components/HomePage';
+import { PeoplePage } from './components/PeoplePage';
+import { PageNotFound } from './components/PageNotFound';
+import { PeopleProvider } from './providers/PeopleProvider';
 
 export const App = () => (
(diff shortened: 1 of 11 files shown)


adrian200002 (approved):
For me it's fine, good job :).

src/App.tsx
```diff
@@ -1,173 +1,29 @@
...
-                  <td>
-                    <a href="#/people/lieven-de-causmaecker-1696">
-                      Lieven de Causmaecker
-                    </a>
-                  </td>
-                </tr>
-              </tbody>
-            </table>
-          </div>
+  <PeopleProvider>
```
danon321, line 10: Does all App need that information or only PeoplePage?

src/components/PeoplePageContent.tsx
```diff
@@ -0,0 +1,32 @@
...
+    }
+
+    return (
+      <p data-cy="noPeopleMessage">
+        There are no people on the server
+      </p>
+    );
+  }
+
+  return null;
```
danon321, line 31: I think better will be to display <ErrorMessage> there than nothing and let user know

src/App.tsx
zaura333: > Does all App need that information or only PeoplePage?

Only PeoplePage and its children components, should I wrap it somewhere else?

PR #1242, opened 2024-01-19
Solution

SashaViun:
[DEMO LINK](https://SashaViun.github.io/react_people-table-basics/)

### DIFF
--- a/src/components/PeopleTable/PeopleTable.tsx
+++ b/src/components/PeopleTable/PeopleTable.tsx
@@ -1,0 +1,82 @@
+import classNames from 'classnames';
+import { useParams, Link } from 'react-router-dom';
+import { Person } from '../../types';
+
+interface PeopleTableProps {
+  people: Person[],
+}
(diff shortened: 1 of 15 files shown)


src/components/PeopleTable/PeopleTable.tsx
```diff
@@ -0,0 +1,82 @@
...
+          <th>Born</th>
+          <th>Died</th>
+          <th>Mother</th>
+          <th>Father</th>
+        </tr>
+      </thead>
+
+      <tbody>
+        {people.map(person => (
+          <>
```
denys-danyliuk, line 30: add key

denys-danyliuk (changes requested):

src/components/PeopleTable/PeopleTable.tsx
```diff
@@ -0,0 +1,85 @@
...
+            <React.Fragment key={person.slug}>
```
etojeDenys, line 32: looks like redundant fragments here

PR #1248, opened 2024-01-23
add task solution

4xmplme:
[DEMO LINK](https://4xmplme.github.io/react_people-table-basics/)


src/api/people.ts
```diff
@@ -7,7 +7,7 @@ function wait(delay: number) {
   return new Promise(resolve => setTimeout(resolve, delay));
 }
 
-export function getPeople(): Promise<Person[]> {
+export async function getPeople(): Promise<Person[]> {
```
DarkMistyRoom, line 10: no need to use async function if we don't use `await`

src/components/PeopleTable.tsx
```diff
@@ -0,0 +1,141 @@
...
+
+      normalizedPerson.motherName = checkName(normalizedPerson.motherName);
+      normalizedPerson.fatherName = checkName(normalizedPerson.fatherName);
+
+      normalizedPerson.mother = findParent(normalizedPerson.motherName);
+      normalizedPerson.father = findParent(normalizedPerson.fatherName);
+
+      return normalizedPerson;
+    })
+  ), []);
```
DarkMistyRoom, line 63: better to move these functions out of the component to utils folder

DarkMistyRoom (approved):
Good job!

PR #1253, opened 2024-01-29
add task solution

7yrak34:
[DEMO LINK](https://7yrak.github.io/react_people-table-basics/)

### DIFF
--- a/src/components/PeopleTable/PeopleTable.tsx
+++ b/src/components/PeopleTable/PeopleTable.tsx
@@ -1,0 +1,99 @@
+/* eslint-disable jsx-a11y/control-has-associated-label */
+import React, { useMemo } from 'react';
+import { useParams } from 'react-router-dom';
+import classNames from 'classnames';
+import { Person } from '../../types';
+import { PersonLink } from '../PersonLink';
+
+interface Props {
+  people: Person[];
+}
+
+export const PeopleTable: React.FC<Props> = ({ people }) => {
+  const preparedPeople = useMemo(() => {
+    return people.map(person => {
+      const father = people.find(p => p.name === person.fatherName);
+      const mother = people.find(p => p.name === person.motherName);
+
+      return { ...person, father, mother };
+    });
+  }, [people]);
+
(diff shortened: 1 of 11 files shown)


src/components/PeopleTable/PeopleTable.tsx
```diff
@@ -0,0 +1,99 @@
...
+                  slug={slug}
+                />
+              </td>
+
+              <td>{sex}</td>
+              <td>{born}</td>
+              <td>{died}</td>
+              <td>
+                {!motherName && '-'}
+                {(motherName && mother) && (
```
witflash, line 73: You can check here just a `mother` because if you have a `mother` you have `motherName` already.

src/components/PeopleTable/PeopleTable.tsx
```diff
@@ -0,0 +1,99 @@
...
+                  <PersonLink
+                    name={motherName}
+                    sex="f"
+                    slug={mother.slug}
+                  />
+                )}
+                {(motherName && !mother) && motherName}
```
witflash, line 80: There is no need to use conditional render.
You can write directly:
```
{motherName}
```
And if `motherName === undefined` (or `null`, an empty string) React renders nothing for this case.

src/components/PeopleTable/PeopleTable.tsx
```diff
@@ -0,0 +1,99 @@
...
+              <td>
+                {!fatherName && '-'}
+                {(fatherName && father) && (
+                  <PersonLink
+                    name={fatherName}
+                    sex="m"
+                    slug={father.slug}
+                  />
+                )}
+                {(fatherName && !father) && fatherName}
```
witflash, line 91: The same approach here (as for `motherName`)

witflash (changes requested):
Good work!
Overall everything is pretty well. But you can improve a bit. 
Check my comments, please.

volodymyr-soltys97 (approved):
Nice work 🔥

PR #1266, opened 2024-02-08
add task solution

rvoshchylo:
[DEMO LINK](https://rvoshchylo.github.io/react_people-table-basics/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,173 +1,34 @@
-import { Loader } from './components/Loader';
+import { Navigate, Route, Routes } from 'react-router-dom';
 
 import './App.scss';
+import { NavBar } from './NavBar';
+import { PeopleTable } from './PeopleTable';
+import { HomePage } from './HomePage';
+import { PageNotFound } from './PageNotFound';
 
-export const App = () => (
-  <div data-cy="app">
-    <nav
-      data-cy="nav"
-      className="navbar is-fixed-top has-shadow"
-      role="navigation"
-      aria-label="main navigation"
-    >
-      <div className="container">
-        <div className="navbar-brand">
-          <a className="navbar-item" href="#/">
-            Home
-          </a>
+export const App = () => {
+  return (
+    <div data-cy="app">
+      <NavBar />
 
-          <a
-            className="navbar-item has-background-grey-lighter"
-            href="#/people"
-          >
-            People
-          </a>
+      <main className="section">
+        <div className="container">
+          <Routes>
+            <Route path="/">
+              <Route index element={<HomePage />} />
+              <Route
+                path="home"
+                element={<Navigate to=".." replace />}
+              />
+              <Route path="people">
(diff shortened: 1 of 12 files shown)


src/App.tsx
```diff
@@ -1,173 +1,34 @@
...
+              <Route
+                path="home"
+                element={<Navigate to=".." replace />}
+              />
+              <Route path="people">
+                <Route path=":slug?" element={<PeoplePage />} />
+              </Route>
+            </Route>
+            <Route path="*" element={<PageNotFound />} />
+          </Routes>
```
artem8746, line 28: would be better to move routes to `Root` component

src/pages/PersonInfo.tsx
```diff
@@ -0,0 +1,61 @@
...
+          sex={sex}
+        />
+      </td>
+
+      <td>{sex}</td>
+      <td>{born}</td>
+      <td>{died}</td>
+      <td>
+        {mother
+          && motherName
```
artem8746, line 49: you don't need this

artem8746 (approved):
Incredible job😻😻

src/App.tsx
olya-shyian: agree, put your `Routes` into component

src/components/Loader/Error.tsx
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+
+interface ErrorMessage {
+  error: string;
+}
+export const Error: React.FC<ErrorMessage> = ({ error }) => {
+  return (
```
olya-shyian, line 7: return jsx without word `return` (if linter allows)
check in all other components

src/pages/PeopleTable.tsx
```diff
@@ -0,0 +1,42 @@
...
+  people: Person[]
+}
+
+export const PeopleTable: React.FC<Props> = ({ people }) => {
+  const normalizedPeople = people.map(person => {
+    return {
+      ...person,
+      father: people.find(father => father.name === person.fatherName),
+      mother: people.find(mother => mother.name === person.motherName),
+    };
```
olya-shyian, line 16: great desicion! 🔥 I like it

also add check `|| null`
so if you dont find any parent its better to have `null` value instead of `undefined`

and you can move this function to `utils` folder

src/pages/PersonInfo.tsx
olya-shyian, line 49: its enough to check if we have `mother` or `father`

olya-shyian (changes requested):
Fantastic! 💫

olya-shyian (approved):
Super!🌟

P.S. If you don't agree with something please explain why =)

rvoshchylo:
> Super!🌟
> 
> P.S. If you don't agree with something please explain why =)

why?) all right, I'm totally agree with u!👻

PR #1303, opened 2024-02-19
solution

mariuszgit:
https://mariuszgit.github.io/react_people-table-basics/

### DIFF
--- a/src/pages/PeoplePage.tsx
+++ b/src/pages/PeoplePage.tsx
@@ -34,9 +34,7 @@ const PeoplePage = () => {
   );
 
   const NoPeopleMsg = () => (
-    <p data-cy="peopleLoadingError" className="has-text-danger">
-      Something went wrong
-    </p>
+    <p data-cy="noPeopleMessage">There are no people on the server</p>
   );
 
   return (


src/components/Table.tsx
```diff
@@ -0,0 +1,85 @@
...
+    >
+      {person.name}
+    </Link>
+  );
+};
+
+const Table = ({ people }: { people: Person[] }) => {
+  const param = useParams();
+
+  const computedPeople = useMemo(
```
yevhenii-pyl, line 21: I have my doubts regarding this `useMemo` usage. 

Sure, it looks like we are making some sort of a complex computation here: map + find + find. 

However, from the other side, you are passing an array of people as a dependancy here. `useMemo` and `useEffect` use shallow comparison to determine if dependencies have changed. If the object or array is recreated on every render (even if its content doesn't change), it will cause the memoized value to be recomputed unnecessarily.

src/components/Table.tsx
```diff
@@ -0,0 +1,85 @@
...
+          <th>Sex</th>
+          <th>Born</th>
+          <th>Died</th>
+          <th>Mother</th>
+          <th>Father</th>
+        </tr>
+      </thead>
+
+      <tbody>
+        {computedPeople.map(person => (
```
yevhenii-pyl, line 49: Let's extract all necessary keys here to avoid typing `person.` each time. 
`({ name, ... }) `

src/components/Table.tsx
```diff
@@ -0,0 +1,85 @@
+/* eslint-disable jsx-a11y/control-has-associated-label */
+import { Link, useParams } from 'react-router-dom';
+import cn from 'classnames';
+import { useMemo } from 'react';
+import { Person } from '../types';
+
+const PersonLink = ({ person }: { person: Person }) => {
```
yevhenii-pyl, line 7: It will be clearer if we just pass necessary props here: slug, sex, name.

src/pages/PeoplePage.tsx
```diff
@@ -0,0 +1,63 @@
...
+    const getData = async () => {
+      setIsLoading(true);
+
+      try {
+        const response = await getPeople();
+
+        setPeople(response);
+      } catch (error) {
+        setIsError(true);
+      } finally {
```
yevhenii-pyl, line 22: GJ, try => catch => finally. Nice and simple.

src/pages/PeoplePage.tsx
```diff
@@ -0,0 +1,63 @@
...
+    getData();
+  }, []);
+
+  return (
+    <>
+      <h1 className="title">People Page</h1>
+
+      <div className="block">
+        <div className="box table-container">
+          {isError ? (
```
yevhenii-pyl, line 36: Could you think of how can you avoid that messy condition in render?

yevhenii-pyl (changes requested):
Solid work, let's just adjust a few things here. Basically, my main concern is that nasty nested ternary at PeoplePage container.

yevhenii-pyl (approved):
Good job 👍 
I would still change the nested ternary at PeoplePage container to at least something like : 

`{isLoading ? ...}`

`{!isLoading && isError ? ...}`

`!isLoading && !isError && !!people.length ? ...  : ...`

Or you can always use this library, it really makes conditional rendering more human readable: https://www.npmjs.com/package/jsx-control-statements

However, since it's more a code styling approach, I see no reason to block the PR. Just keep in mind that nested ternaries may become (like, REALLY quick) painful and hard to manage.

PR #1383, opened 2024-04-09
add solution

tetlisna:
[DEMO LINK](https://tetlisna.github.io/react_people-table-basics/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,167 +1,38 @@
-import { Loader } from './components/Loader';
-
 import './App.scss';
 
-export const App = () => (
-  <div data-cy="app">
-    <nav
-      data-cy="nav"
-      className="navbar is-fixed-top has-shadow"
-      role="navigation"
-      aria-label="main navigation"
-    >
-      <div className="container">
-        <div className="navbar-brand">
-          <a className="navbar-item" href="#/">
-            Home
-          </a>
+import { Navigate, Route, Routes } from 'react-router-dom';
 
-          <a
-            className="navbar-item has-background-grey-lighter"
-            href="#/people"
-          >
-            People
-          </a>
+import { NavBar } from './components/NavBar/NavBar';
+import { PeoplePage } from './components/PeoplePage/PeoplePage';
+import { HomePage } from './components/HomePage/HomePage';
+import { NotFoundPage } from './components/NotFoundPage/NotFoundPage';
+
+export const App = () => {
+  return (
+    <div data-cy="app">
+      <nav
+        data-cy="nav"
+        className="navbar is-fixed-top has-shadow"
+        role="navigation"
+        aria-label="main navigation"
+      >
+        <div className="container">
+          <NavBar />
         </div>
-      </div>
-    </nav>
+      </nav>
 
-    <main className="section">
-      <div className="container">
-        <h1 className="title">Home Page</h1>
-        <h1 className="title">People Page</h1>
-        <h1 className="title">Page not found</h1>
-
-        <div className="block">
-          <div className="box table-container">
-            <Loader />
-
-            <p data-cy="peopleLoadingError" className="has-text-danger">
-              Something went wrong
-            </p>
-
-            <p data-cy="noPeopleMessage">There are no people on the server</p>
-
-            <table
-              data-cy="peopleTable"
-              className="table is-striped is-hoverable is-narrow is-fullwidth"
-            >
-              <thead>
-                <tr>
-                  <th>Name</th>
-                  <th>Sex</th>
-                  <th>Born</th>
-                  <th>Died</th>
-                  <th>Mother</th>
-                  <th>Father</th>
-                </tr>
-              </thead>
-
-              <tbody>
-                <tr data-cy="person">
-                  <td>
-                    <a href="#/people/jan-van-brussel-1714">Jan van Brussel</a>
-                  </td>
-
-                  <td>m</td>
-                  <td>1714</td>
-                  <td>1748</td>
-                  <td>Joanna van Rooten</td>
-                  <td>Jacobus van Brussel</td>
-                </tr>
-
-                <tr data-cy="person">
-                  <td>
-                    <a href="#/people/philibert-haverbeke-1907">
-                      Philibert Haverbeke
-                    </a>
-                  </td>
-
-                  <td>m</td>
-                  <td>1907</td>
-                  <td>1997</td>
-
-                  <td>
-                    <a
-                      className="has-text-danger"
-                      href="#/people/emma-de-milliano-1876"
-                    >
-                      Emma de Milliano
-                    </a>
-                  </td>
-
-                  <td>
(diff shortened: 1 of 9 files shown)


src/components/PeopleTable/PeopleTable.tsx
```diff
@@ -0,0 +1,100 @@
...
+                  <td>{died}</td>
+
+                  <td>
+                    {mother && (
+                      <Link
+                        className="has-text-danger"
+                        to={`/people/${mother.slug}`}
+                      >
+                        {motherName}
+                      </Link>
```
vlesia, line 81: It would be better to create a ` <PersonLink person={person} />` then you can use it for 
`<PersonLink person={mother} /> and <PersonLink person={farther} />`

vlesia (approved):
Great job :fire:

src/App.tsx
```diff
@@ -1,167 +1,38 @@
...
+      <main className="section">
+        <div className="container">
+          <Routes>
+            <Route path="/" element={<HomePage />} />
+            <Route path="home" element={<Navigate to="/" replace />} />
+            <Route path="/people" element={<PeoplePage />}>
+              <Route path=":path" element={<PeoplePage />} />
+            </Route>
+            <Route path="*" element={<NotFoundPage />} />
+          </Routes>
```
dffUqp, line 33: if I decided to add a new page which will not include a `nav`, how can I do this?

I suggest you create a `Layout` component and wrap all your routes in the `Layout` route.

```jsx
<Routes>
       <Route path="/" element={<Layout/>}>
             ...
        </Route>
</Routes>
```
In Layout, you can use `Outlet` from `react-router-dom` to render your components inside.

src/components/NavBar/NavBar.tsx
```diff
@@ -0,0 +1,21 @@
+import cn from 'classnames';
+import { NavLink } from 'react-router-dom';
+
+export const NavBar = () => {
+  const getIsActive = ({ isActive }: { isActive: boolean }) =>
+    cn('navbar-item', {
+      'has-background-grey-lighter': isActive,
+    });
```
dffUqp, line 8: I think it will be better to move this function outside our component. 

Currently, we create a new function on each `NavBar` re-render.

src/components/PeoplePage/PeoplePage.tsx
```diff
@@ -0,0 +1,29 @@
...
+        </p>
+      ) : (
+        <>
+          {people.length ? (
+            <PeopleTable />
+          ) : (
+            <p data-cy="noPeopleMessage">There are no people on the server</p>
+          )}
+        </>
+      )}
```
dffUqp, line 26: Nested ternary operator is bad practice

src/components/PeopleTable/PeopleTable.tsx
```diff
@@ -0,0 +1,38 @@
+import { PersonLink } from '../PersonLink/PersonLink';
+import { usePeople } from '../context/PeopleContext';
+
+export const PeopleTable: React.FC = () => {
+  const columns = ['Name', 'Sex', 'Born', 'Died', 'Mother', 'Father'];
```
dffUqp, line 5: Move it outside of the component

src/components/PersonLink/PersonLink.tsx
```diff
@@ -0,0 +1,65 @@
...
+    born,
+    died,
+    fatherName,
+    motherName,
+    slug,
+    mother,
+    father,
+  } = person;
+
+  const male = sex === 'm';
```
dffUqp, line 25: 1. let's create variable outside component for "m" because it's magic value
```jsx
const MALE = 'm';
```

2. Naming convention can be different from project to project but in general, we name boolean variables with prefix `is` or `has`

```jsx
const isMale = sex === MALE;
```

src/components/PersonLink/PersonLink.tsx
```diff
@@ -0,0 +1,65 @@
...
+      <td>{born}</td>
+      <td>{died}</td>
+
+      <td>
+        {mother && (
+          <Link className="has-text-danger" to={`/people/${mother.slug}`}>
+            {motherName}
+          </Link>
+        )}
+        {!mother && (motherName || '-')}
```
dffUqp, line 56: `-` is magic value

src/components/PersonLink/PersonLink.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+      <td>{sex}</td>
+      <td>{born}</td>
+      <td>{died}</td>
+
+      <td>
+        {mother && (
+          <Link className="has-text-danger" to={`/people/${mother.slug}`}>
+            {motherName}
+          </Link>
```
dffUqp, line 54: Create the `<PersonLink person={person} />` and use it for all existing people

src/components/context/PeopleContext.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+  useEffect(() => {
+    setIsLoading(true);
+
+    const fetchPeople = async () => {
+      try {
+        const fetchedPeople = await getPeople();
+
+        setPeople(fetchedPeople);
+      } catch {
```
dffUqp, line 37: In the catch block, you should also catch a possible error, because now the error does not look informative, for example, you can do this:

```js

catch(error) { setIsError(`Something went wrong: ${error.message}`)}
```

src/components/context/PeopleContext.tsx
```diff
@@ -0,0 +1,62 @@
...
+      {children}
+    </PeopleContext.Provider>
+  );
+};
+
+export const usePeople = () => {
+  const context = useContext(PeopleContext);
+
+  if (!context) {
+    throw new Error('error');
```
dffUqp, line 58: I think there is not enough information about what this error is 😶‍🌫️

src/components/PersonLink/PersonLink.tsx
```diff
@@ -0,0 +1,65 @@
...
+  } = person;
+
+  const male = sex === 'm';
+
+  return (
+    <tr
+      data-cy="person"
+      key={slug}
+      className={cn({
+        'has-background-warning': slug === path,
```
dffUqp, line 32: Let's create a separate variable for `slug === path`

dffUqp (changes requested):
Almost done🙂

BudnikOleksii (changes requested):
For some reason, it's in the queue for review, but Yurii has already left the suggestions

src/components/PeopleTable/PeopleTable.tsx
```diff
@@ -0,0 +1,79 @@
...
+                mother,
+                father,
+              } = person;
+
+              const noParents = '-';
+              const isClickedPerson = path === slug;
+
+              return (
+                <tr
+                  data-cy="person"
```
BudnikOleksii, line 51: Create a person component

BudnikOleksii (approved):
Good job 👍

PR #1406, opened 2024-04-25
solution

KachVl:
[DEMO](https://kachvl.github.io/react_people-table-basics/)

### DIFF
--- a/src/components/PersonLink.tsx
+++ b/src/components/PersonLink.tsx
@@ -1,14 +1,12 @@
-import classNames from "classnames";
-import { Person } from "../types";
-import { Link, useParams } from "react-router-dom";
-import { useContext } from "react";
-import { PeopleContext } from "./PeopleContext";
+import classNames from 'classnames';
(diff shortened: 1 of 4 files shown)


src/components/PersonLink.tsx
```diff
@@ -0,0 +1,72 @@
...
+import { PeopleContext } from './PeopleContext';
+
+type Props = {
+  person: Person;
+};
+
+export const PersonLink: React.FC<Props> = ({ person }) => {
+  const { name, sex, born, died, fatherName, motherName } = person;
+  const { slug } = useParams();
+  const { peopleList } = useContext(PeopleContext);
```
TarasHoliuk, line 14: As far as I see you use `peopleList` here and in the PeoplePage component, where you fetch and set it. So in this case without context, you would need to pass `peopleList` as prop only to 1 component. I.e. Context is redundant here.

Just FYI

TarasHoliuk (approved):
GJ 👍 Approved

PR #1421, opened 2024-05-08
Develop

yurovych:
https://yurovych.github.io/react_people-table-basics/


DarkMistyRoom (approved):
Good job! But usually, we name page and component files with UpperCase

PR #1430, opened 2024-05-15
Develop

marinatea:
[DEMO LINK](https://marinatea.github.io/react_people-table-basics/)


src/components/NavBar.tsx
```diff
@@ -0,0 +1,52 @@
...
+    } else if (currentLink.startsWith('/people')) {
+      setIsActive('people');
+    } else {
+      setIsActive('');
+    }
+  }, [location]);
+
+  const handleClick = (link: string) => {
+    setIsActive(link);
+  };
```
nataliaklonowska, line 22: it is a creative way to check if link is active but in NavLink you can check if it's active ;) https://reactrouter.com/en/main/components/nav-link
Try to use this built-in feature instead of creating state isActive

nataliaklonowska (changes requested):

choeqq (approved):

PR #1440, opened 2024-05-19
solution react people table basics

andreyysak:
[DEMO LINK](https://andreyysak.github.io/react_people-table-basics/)


vadiimvooo (changes requested):
Almost done, but there is an interesting bug you have to fix.
https://www.loom.com/share/181e4188cce343909b979149b199b51b

PR #1441, opened 2024-05-20
Solution

Allanpt:
[DEMO LINK](https://Allanpt.github.io/react_people-table-basics/)


jv-aquino (changes requested):
Clean code Allan! Nice! I think there's a problem with the your tests though

\+ a quick suggestion (totally optional):
Try to create a smooth scroll effect leading to where the mother/father is on the table
image

jv-aquino (approved):
Great job Allan! I was talking about like using ids and <a> elements to direct the user to the row, but no problem, it was just a suggestion, let's focus on the more advanced stuff

PR #1448, opened 2024-05-23
task solution

MatheusStefan:
[DEMO LINK](https://MatheusStefan.github.io/react_people-table-basics/)

### DIFF
--- a/src/pages/people-page/people.page.tsx
+++ b/src/pages/people-page/people.page.tsx
@@ -1,0 +1,123 @@
+import { useEffect, useState } from 'react';
+import { Loader } from '../../components/Loader';
+import { getPeople } from '../../api';
+import { Person } from './../../types/Person';
+import { PersonLink } from '../person-link/person.link';
+import { useParams } from 'react-router-dom';
+
+export default function PeoplePage() {
+  const [isLoading, setIsLoading] = useState(true);
+  const [people, setPeople] = useState<Person[]>([]);
+  const [selectedPersonSlug, setSelectedPersonSlug] = useState<string | null>(
+    null,
+  );
+  const [error, setError] = useState<string | null>(null);
+  const { slug: targetPersonSlug } = useParams<{ slug: string }>();
+
+  useEffect(() => {
+    getPeople()
+      .then(fetchedPeople => {
+        setPeople(fetchedPeople);
+        setIsLoading(false);
+        setError(null);
+      })
+      .catch(() => {
+        setIsLoading(false);
+        setError('Something went wrong');
+      })
+      .finally(() => {
+        setIsLoading(false);
+      });
+  }, []);
+
+  const handlePersonClick = (slug: string | null) => {
+    setSelectedPersonSlug(slug);
+  };
+
+  const shouldApplyBackground = (slug: string) =>
+    targetPersonSlug === slug ? 'has-background-warning' : '';
+
+  return (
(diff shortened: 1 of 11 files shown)


src/pages/people-page/people.page.tsx
```diff
@@ -0,0 +1,123 @@
...
+                  >
+                    <td className={shouldApplyBackground(person.slug)}>
+                      <PersonLink
+                        person={person}
+                        onClick={() => handlePersonClick(person.slug)}
+                      >
+                        {person.name}
+                      </PersonLink>
+                    </td>
+                    <td className={shouldApplyBackground(person.slug)}>
```
pedro-ruas, line 83: This code looks way too repetitive, there must be a way to make it simpler...
What happens if all _td_ used a _PersonLink_ inside? Maybe you could use a _map_:
```
const tableCells = ['name', 'sex', 'born', 'died', 'motherName', 'fatherName'];

return (
...
{people.map(person => (
    <tr
        key={person.slug}
        data-cy="person"
        className={shouldApplyBackground(person.slug)}
    >
        {tableCells.map(cell => (
            <td>
                <PersonLink person={peopleMap[person[cell]]}>
                   {person[cell] !== undefined ? person[cell] : '-'}
               </PersonLink>
           </td>
......
```

src/pages/person-link/person.link.tsx
```diff
@@ -0,0 +1,29 @@
...
+export const PersonLink: React.FC<
+PersonLinkProps & { onClick?: () => void }
+> = ({ person, children, onClick }) => {
+  const handleClick = () => {
+    if (onClick) {
+      onClick();
+    }
+  };
+
+  const foundPerson = person ? person : null;
```
pedro-ruas, line 14: You don't need _foundPerson_, you could do a conditional rendering as in the previous files:
```
return (
    <>
        {person ? (
           <Link ...
         ) : (
           <>{children}</>
    </>
)
```

src/types/Person.link.ts
```diff
@@ -0,0 +1,6 @@
+import { Person } from './Person';
+
+export type PersonLinkProps = {
+  person: Person | undefined;
```
pedro-ruas, line 4: Change this to:
`
person?: Person `;

src/pages/people-page/people.page.tsx
```diff
@@ -0,0 +1,123 @@
...
+                    </td>
+                    <td
+                      className={
+                        selectedPersonSlug === person.slug
+                          ? 'has-background-warning'
+                          : ''
+                      }
+                    >
+                      <PersonLink
+                        person={people.find(p => p.name === person.fatherName)}
```
pedro-ruas, line 108: This works just fine, but there is another approach that might be better in performance:
Instead of using _find_ method in all PersonLinks, which iterates the whole array until it finds a result, you could use a single iteration using a _.reduce_ method and creating either a _Map_ or an _Object_ to access it by keys, e.g.:
```
const peopleMap = people.reduce((map, person) => ({ ...map, [person.name]: person }), {});
...
<PersonLink person={peopleMap[p.fatherName]}
```
This way, you reduce N iterations in _people_ array into a single iteration (reduce method iteration) and then access the parents by using their hash, which is much more performatic than using a _.find_

jv-aquino (approved):
nice Matheus, great job with the corrections!

PR #1454, opened 2024-05-27
Solution

odunchyk:
[DEMO LINK](https://odunchyk.github.io/react_people-table-basics/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,9 @@
-import { Loader } from './components/Loader';
-
 import './App.scss';
+import { NavLink, Outlet } from 'react-router-dom';
+import classNames from 'classnames';
+
+const getLinkClass = ({ isActive }: { isActive: boolean }) =>
+  classNames('navbar-item', { 'has-background-grey-lighter': isActive });
 
 export const App = () => (
   <div data-cy="app">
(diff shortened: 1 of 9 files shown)


src/pages/PeoplePage.tsx
```diff
@@ -0,0 +1,121 @@
...
+                  <p data-cy="peopleLoadingError" className="has-text-danger">
+                    {errorMessage}
+                  </p>
+                )}
+
+                {people?.length === 0 ? (
+                  <p data-cy="noPeopleMessage">
+                    There are no people on the server
+                  </p>
+                ) : (
```
luyx0, line 47: I think it is better to do it not through the ternary operator, but simply through if

src/pages/PeoplePage.tsx
```diff
@@ -0,0 +1,121 @@
...
+    setIsLoading(true);
+
+    getPeople()
+      .then(setPeople)
+      .catch(() => setErrorMessage('Something went wrong'))
+      .finally(() => setIsLoading(false));
+  }, []);
+  return (
+    <>
+      <h1 className="title">People Page</h1>
```
luyx0, line 28: add upper `<div className="block">`

src/index.tsx
```diff
@@ -1,13 +1,31 @@
...
-    <App />
+    <Routes>
+      <Route path="/" element={<App />}>
+        <Route index element={<HomePage />} />
+        <Route path="home" element={<Navigate to="/" replace />} />
+        <Route path="people">
+          <Route index element={<PeoplePage />} />
+          <Route path=":personSlug?" element={<PeoplePage />} />
+        </Route>
+        <Route path="*" element={<PageNotFound />} />
```
luyx0, line 27: You refactor this code on this: `
<Routes>
            <Route path="/" element={<HomePage />} />
            <Route path="/home" element={<Navigate to="/" replace />} />
            <Route path="/people" element={<PeoplePage />} />
            <Route path="/people/:slug" element={<PeoplePage />} />
            <Route path="*" element={<NotFoundPage />} />
          </Routes>
`

luyx0 (changes requested):
The code looks good, but there are many small bugs that need to be fixed, I also don't like the fact that Routes are found in the index

src/pages/PeoplePage.tsx
odunchyk: Yeah, you right. PeoplePage need refactoring - too hard to read right now

src/pages/PeoplePage.tsx
odunchyk: Why?
JSX fragment doing the same job here

src/index.tsx
odunchyk: My routes have hierarchy
i think it is more readable like i did

src/App.tsx
```diff
@@ -11,157 +14,20 @@ export const App = () => (
       aria-label="main navigation"
     >
       <div className="container">
-        <div className="navbar-brand">
-          <a className="navbar-item" href="#/">
+        <div className={'navbar-brand'}>
+          <NavLink className={getLinkClass} to="/">
```
DanilWeda, line 18: You can move all paths in enums or constants and use it in all places in your application.

src/index.tsx
DanilWeda: Not critical

DanilWeda (approved):
Good job! 
Approved with small comments. 
Thanks)

PR #1457, opened 2024-05-27
add task solution

PrytulaAlexandr:
[DEMO LINK](https://PrytulaAlexandr.github.io/react_people-table-basics/)

### DIFF
--- a/src/Root.tsx
+++ b/src/Root.tsx
@@ -1,0 +1,25 @@
+import {
+  Navigate,
+  Route,
+  HashRouter as Router,
+  Routes,
+} from 'react-router-dom';
+import { App } from './App';
+import HomePage from './pages/HomePage';
+import PeoplePage from './pages/PeoplePage';
+import PageNotFound from './pages/PageNotFound';
+
+export const Root = () => (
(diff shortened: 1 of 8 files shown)


PrytulaAlexandr:
I don't understand, all the tests are running on my local machine, but here for some reason the linter doesn't like something.

src/Root.tsx
```diff
@@ -0,0 +1,25 @@
...
+} from 'react-router-dom';
+import { App } from './App';
+import HomePage from './pages/HomePage';
+import PeoplePage from './pages/PeoplePage';
+import PageNotFound from './pages/PageNotFound';
+
+export const Root = () => (
+  <Router>
+    <Routes>
+      <Route path="/" element={<App />}>
```
DanilWeda, line 15: You can move all paths in enum or constants and use them in all places in your application.

DanilWeda (approved):
Great. 
Approved! 
Thanks)

src/Root.tsx
PrytulaAlexandr: Danil thanks for check.

PR #1482, opened 2024-05-31
add task solution

SlawikH:
[DEMO LINK](https://SlawikH.github.io/react_people-table-basics/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,167 +1,24 @@
-import { Loader } from './components/Loader';
-
 import './App.scss';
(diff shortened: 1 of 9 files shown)


src/App.tsx
```diff
@@ -1,167 +1,24 @@
-import { Loader } from './components/Loader';
-
 import './App.scss';
-
```
sevenlays, line 4: :(

sevenlays (changes requested):

PR #1494, opened 2024-06-17
solution

Ehulinian:
[DEMO LINK](https://Ehulinian.github.io/react_people-table-basics/) and add it to the PR description.

### DIFF
--- a/src/components/PeopleTable.tsx
+++ b/src/components/PeopleTable.tsx
@@ -1,0 +1,68 @@
+import { FC } from 'react';
+
+import { Person } from '../types';
+import PersonLink from './PersonLink';
+
+interface Props {
+  people: Person[];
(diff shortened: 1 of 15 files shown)


src/components/PeopleTable.tsx
```diff
@@ -0,0 +1,68 @@
...
+      </thead>
+
+      <tbody>
+        {people.map(person => (
+          <tr
+            key={person.slug}
+            data-cy="person"
+            className={
+              selectedSlug === person.slug ? 'has-background-warning' : ''
+            }
```
SanyaBratashchuk, line 49: It is better to use libraries like classnames or clsx for conditional classnames

SanyaBratashchuk (approved):

PR #1508, opened 2024-07-02
task

rwlp:
[DEMO LINK](https://rwlp.github.io/react_people-table-basics/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,167 +1,17 @@
-import { Loader } from './components/Loader';
-
 import './App.scss';
+import { NavBar } from './components/NavBar/NavBar';
+import { Router } from './components/router/router';
 
 export const App = () => (
-  <div data-cy="app">
-    <nav
-      data-cy="nav"
-      className="navbar is-fixed-top has-shadow"
-      role="navigation"
-      aria-label="main navigation"
-    >
-      <div className="container">
-        <div className="navbar-brand">
-          <a className="navbar-item" href="#/">
-            Home
-          </a>
+  <>
+    <div data-cy="app">
+      <NavBar />
 
-          <a
-            className="navbar-item has-background-grey-lighter"
-            href="#/people"
-          >
-            People
-          </a>
+      <main className="section">
+        <div className="container">
+          <Router />
         </div>
-      </div>
-    </nav>
-
-    <main className="section">
-      <div className="container">
-        <h1 className="title">Home Page</h1>
-        <h1 className="title">People Page</h1>
-        <h1 className="title">Page not found</h1>
-
-        <div className="block">
-          <div className="box table-container">
-            <Loader />
-
-            <p data-cy="peopleLoadingError" className="has-text-danger">
-              Something went wrong
-            </p>
-
-            <p data-cy="noPeopleMessage">There are no people on the server</p>
-
-            <table
-              data-cy="peopleTable"
-              className="table is-striped is-hoverable is-narrow is-fullwidth"
-            >
-              <thead>
-                <tr>
-                  <th>Name</th>
-                  <th>Sex</th>
-                  <th>Born</th>
-                  <th>Died</th>
-                  <th>Mother</th>
-                  <th>Father</th>
-                </tr>
-              </thead>
-
-              <tbody>
-                <tr data-cy="person">
-                  <td>
-                    <a href="#/people/jan-van-brussel-1714">Jan van Brussel</a>
-                  </td>
-
-                  <td>m</td>
-                  <td>1714</td>
-                  <td>1748</td>
-                  <td>Joanna van Rooten</td>
-                  <td>Jacobus van Brussel</td>
-                </tr>
-
-                <tr data-cy="person">
-                  <td>
-                    <a href="#/people/philibert-haverbeke-1907">
-                      Philibert Haverbeke
-                    </a>
-                  </td>
-
-                  <td>m</td>
-                  <td>1907</td>
-                  <td>1997</td>
-
-                  <td>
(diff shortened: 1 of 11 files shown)


src/App.tsx
```diff
@@ -1,167 +1,17 @@
...
-
-                  <td>
-                    <a
-                      className="has-text-danger"
-                      href="#/people/livina-haverbeke-1692"
-                    >
-                      Livina Haverbeke
-                    </a>
-                  </td>
+  <>
```
pedro-ruas, line 6: Do you need this _Fragment_?

src/components/Table/Table.tsx
```diff
@@ -0,0 +1,142 @@
...
+export const Table: React.FC = () => {
+  const { personSlug } = useParams();
+  const [requestState, setRequestState] = useState('loading');
+  const [peopleTable, setPeopleTable] = useState<Person[] | null>(null);
+
+  useEffect(() => {
+    fetch('https://mate-academy.github.io/react_people-table/api/people.json');
+    getPeople()
+      .then((data: Person[]) => {
+        if (data.length === 0) {
```
pedro-ruas, line 16: Why are you checking this?
If the array is empty, `treatedData` will also be an empty array, the _.map_ method will not run on an empty array...

src/components/Table/Table.tsx
```diff
@@ -0,0 +1,142 @@
...
+          <Loader />
+        ) : (
+          <>
+            {requestState.includes('error') && (
+              <>
+                <p data-cy="peopleLoadingError" className="has-text-danger">
+                  {' '}
+                  Something went wrong
+                </p>
+                {requestState.includes('empty') && (
```
pedro-ruas, line 50: I which case will this line run?

src/components/Table/Table.tsx
```diff
@@ -0,0 +1,142 @@
+import React, { useState, useEffect } from 'react';
+import { Person } from '../../types';
+import { useParams } from 'react-router-dom';
+import { Loader } from '../Loader';
+import { getPeople } from '../../api';
+
+export const Table: React.FC = () => {
+  const { personSlug } = useParams();
+  const [requestState, setRequestState] = useState('loading');
+  const [peopleTable, setPeopleTable] = useState<Person[] | null>(null);
```
pedro-ruas, line 10: I don't see the point of passing a _null_ value here instead of simply using an empty array as initial value

src/components/Table/Table.tsx
```diff
@@ -0,0 +1,142 @@
+import React, { useState, useEffect } from 'react';
+import { Person } from '../../types';
+import { useParams } from 'react-router-dom';
+import { Loader } from '../Loader';
+import { getPeople } from '../../api';
+
+export const Table: React.FC = () => {
+  const { personSlug } = useParams();
+  const [requestState, setRequestState] = useState('loading');
```
pedro-ruas, line 9: It's not a great idea to store hard-coded strings like this.
I understand that you wanted to avoid multiple _useState_ declarations and decided to track _errors_, _loading_ and _empty_ in the same state.
However, it really isn't necessary as there is no need to store _empty_ in the state.
The second point is, if you really want to keep your state like this, please move these strings to an _Enum_, to guarantee it will be restricted to the possible cases

src/components/Table/Table.tsx
```diff
@@ -0,0 +1,142 @@
...
+        const treatedData = data.map(e => {
+          const motherSlug = data.find(el => el.name === e.motherName)?.slug;
+          const fatherSlug = data.find(el => el.name === e.fatherName)?.slug;
+
+          return { ...e, motherSlug: motherSlug, fatherSlug: fatherSlug };
+        });
+
+        return treatedData;
+      })
+      .then(data => {
```
pedro-ruas, line 28: No problem, but so you know, you don't need multiple _.then_ calls here

src/components/Table/Table.tsx
```diff
@@ -0,0 +1,142 @@
...
+      })
+      .catch(() => {
+        setRequestState('error');
+      });
+  }, []);
+
+  return (
+    <div className="block">
+      <div className="box table-container">
+        {requestState.includes('loading') ? (
```
pedro-ruas, line 40: Change these hard-coded strings

src/components/Table/Table.tsx
```diff
@@ -0,0 +1,142 @@
...
+                  <th>Sex</th>
+                  <th>Born</th>
+                  <th>Died</th>
+                  <th>Mother</th>
+                  <th>Father</th>
+                </tr>
+              </thead>
+
+              <tbody>
+                {peopleTable?.map(
```
pedro-ruas, line 82: No point in checking for null here...

pedro-ruas (changes requested):

src/components/Table/Table.tsx
rwlp: I think that is about tests applied. But I'll check this again.

src/components/Table/Table.tsx
rwlp: This is becaouse this state can be null and tsc rules and lint force this. This is about your code style whe create a state and don't use null. I'll change this accordly your code style and recomendations.

src/App.tsx
rwlp: In fact I think not.

src/components/Table/Table.tsx
rwlp: I think a diff way bcz tests of mateacademy scripts but I'll work for put your style code here, is possible to change and improve this accordly your style code and advices.

PR #1511, opened 2024-07-04
add task solution

Esther-Guimaraes:
[DEMO LINK](https://Esther-Guimaraes.github.io/react_people-table-basics/)


src/pages/people-page/PeoplePage.tsx
```diff
@@ -0,0 +1,118 @@
...
+
+export const PeoplePage = () => {
+  const [peopleList, setPeopleList] = useState<Person[]>([]);
+  const [hasError, setHasError] = useState(false);
+  const [isLoading, setIsLoading] = useState(false);
+
+  const { peopleSlug } = useParams();
+
+  useEffect(() => {
+    setIsLoading(true);
```
pedro-ruas, line 15: Instead of using setIsLoading(true) here, you could make the initial state as _true_ and skip this setState action

src/pages/people-page/PeoplePage.tsx
```diff
@@ -0,0 +1,118 @@
...
+
+    getPeople()
+      .then(res => {
+        const peopleRes: Person[] = [];
+
+        res.map(person => {
```
pedro-ruas, line 21: _Array.map_ is used when you want the resulting array, your implementation doesn't do it.
There are two approaches you could take to fix it:
1. Use _Array.map_ directly on peopleRes:
```
const peopleRes: Person[] = res.map(person => {
    ...code goes here
    return { ...person, mother, father }
}
```
2. Keep the code as is, but use _Array.forEach_ instead

The first approach is better as you don't need to create a second array (peopleRes) and push to it.

src/pages/people-page/PeoplePage.tsx
```diff
@@ -0,0 +1,118 @@
...
+          const mother = res.find(p => p.name === person.motherName);
+          const father = res.find(p => p.name === person.fatherName);
+
+          peopleRes.push({ ...person, mother: mother, father: father });
```
pedro-ruas, line 25: When declaring an object like this, based on other consts, if the obejct's property has the same name as the const, you can pass only the const, without the property declaration. 
What I mean is that the following code line does the same as your code:
```
peopleRes.push({ ...person, mother, father });
```
Notice how "father: ...", and "mother: ..." can be omitted

src/pages/people-page/PeoplePage.tsx
```diff
@@ -0,0 +1,118 @@
...
+            <p data-cy="peopleLoadingError" className="has-text-danger">
+              Something went wrong
+            </p>
+          )}
+
+          {peopleList.length === 0 && !isLoading && (
+            <p data-cy="noPeopleMessage">There are no people on the server</p>
+          )}
+
+          {peopleList.length > 0 && (
```
pedro-ruas, line 50: All there _if_ statements could be changed to a chained ternary to prevent case overlapping:

```
{isLoading ? (
    <Loader
) : hasError ? (
   ...
) : peopleList.length === 0 ? (
   ...
) : (
   <table ...
)}
```

src/types/Person.ts
```diff
@@ -1,4 +1,6 @@
 export interface Person {
+  setFather(father: Person | undefined): void;
+  setMother(mother: Person | undefined): void;
```
pedro-ruas, line 3: Are these new properties being used?

pedro-ruas (changes requested):

src/pages/page-not-found/PageNotFound.tsx
```diff
@@ -0,0 +1,3 @@
+export const PageNotFound = () => {
+  return <h1 className="title">Page not found</h1>;
+};
```
jv-aquino, line 3: nice!

jv-aquino (approved):
Great job Esther!

PR #1525, opened 2024-07-13
people table basics

robsza1122:
https://robsza1122.github.io/react_people-table-basics/

### DIFF
--- a/src/components/PersonList.tsx
+++ b/src/components/PersonList.tsx
@@ -1,0 +1,128 @@
+import { useEffect, useState } from 'react';
+import { Loader } from './Loader';
+import { Person } from '../types';
+import { getPeople } from '../api';
+import classNames from 'classnames';
+import { NavLink, useParams } from 'react-router-dom';
+
+export const PersonList = () => {
+  const [peopleFromApi, setPeopleFromApi] = useState<Person[]>([]);
+  const [error, setError] = useState(false);
+  const [loading, setLoading] = useState(false);
+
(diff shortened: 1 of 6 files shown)


src/components/PersonList.tsx
```diff
@@ -0,0 +1,128 @@
...
+import { NavLink, useParams } from 'react-router-dom';
+
+export const PersonList = () => {
+  const [peopleFromApi, setPeopleFromApi] = useState<Person[]>([]);
+  const [error, setError] = useState(false);
+  const [loading, setLoading] = useState(false);
+
+  const { slug } = useParams();
+
+  const selectedPerson = slug ? slug : '';
```
natalia-klonowska, line 15: you can shorten code to
```suggestion
  const selectedPerson = slug || '';
```

src/components/PersonList.tsx
```diff
@@ -0,0 +1,128 @@
...
+              </p>
+            )}
+
+            {!loading && !peopleFromApi.length && (
+              <p data-cy="noPeopleMessage">There are no people on the server</p>
+            )}
+            {loading ? (
+              <Loader />
+            ) : (
+              <table
```
natalia-klonowska, line 58: if error occurs there shouldn't be message about no people on server and table should appear only when there is some data in it
image

natalia-klonowska (changes requested):
Great work but you need to better handle error messages ;)

BogdanMaliuta (approved):
GJ!

PR #1570, opened 2024-08-06
create a people-table with help  react_routes

Solovei-Andrii:
Solovei-Andrii>.github.io/react_people-table-basics/

### DIFF
--- a/src/components/People.tsx
+++ b/src/components/People.tsx
@@ -1,0 +1,74 @@
+import { useEffect, useState } from 'react';
+import { getPeople } from '../api';
+import { Person } from '../types';
+import { Loader } from './Loader';
+import { PersonLink } from './PersonLink';
+import { getPeopleWithParents } from '../utils/utils';
+
+export const People: React.FC = () => {
+  const [peopleFromServer, setPeopleFromServer] = useState<Person[]>([]);
+  const [error, setError] = useState<string | null>(null);
+
+  useEffect(() => {
+    const fetchPeople = async () => {
+      try {
+        const data = await getPeople();
+
+        if (Array.isArray(data)) {
+          setPeopleFromServer(getPeopleWithParents(data));
+        }
+      } catch {
+        setError('ERROR');
+      }
+    };
+
+    fetchPeople();
+  }, []);
+
+  return (
+    <>
+      <h1 className="title">People Page</h1>
+
+      <div className="block">
+        <div className="box table-container">
+          {peopleFromServer.length === 0 && error === null && <Loader />}
+
+          {error === 'ERROR' && (
+            <>
+              <p data-cy="peopleLoadingError" className="has-text-danger">
+                Something went wrong
+              </p>
+
+              <p data-cy="noPeopleMessage">There are no people on the server</p>
+            </>
+          )}
+
+          {peopleFromServer.length > 0 && (
+            <table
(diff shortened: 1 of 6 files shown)


anastasiiavorobiova (changes requested):
Great work! Feel free to ask for help in the chat to fix the last test 
Screenshot 2024-08-06 at 14 56 09
And please, add the actual demo link to the PR description https://solovei-andrii.github.io/react_people-table-basics/

src/components/People.tsx
```diff
@@ -0,0 +1,84 @@
...
+                  <th>Sex</th>
+                  <th>Born</th>
+                  <th>Died</th>
+                  <th>Mother</th>
+                  <th>Father</th>
+                </tr>
+              </thead>
+
+              <tbody>
+                {Array.isArray(peopleFromServer) &&
```
volodymyr-soltys97, line 63: Move this logic `Array.isArray(peopleFromServer)` to the helper variable and use it here
```suggestion
                {Array.isArray(peopleFromServer) &&
```

src/components/People.tsx
```diff
@@ -0,0 +1,88 @@
...
+import { getPeople } from '../api';
+import { Person } from '../types';
+import { Loader } from './Loader';
+import { PersonLink } from './PersonLink';
+import { getPeopleWithParents } from '../utils/utils';
+
+export const People: React.FC = () => {
+  const [peopleFromServer, setPeopleFromServer] = useState<Person[] | null>(
+    null,
+  );
```
etojeDenys, line 11:
```suggestion
  const [peopleFromServer, setPeopleFromServer] = useState<Person[]>([]);
```
you can simplify it

src/components/People.tsx
```diff
@@ -0,0 +1,88 @@
...
+
+          {error === 'ERROR' && (
+            <>
+              <p data-cy="peopleLoadingError" className="has-text-danger">
+                Something went wrong
+              </p>
+
+              <p data-cy="noPeopleMessage">There are no people on the server</p>
+            </>
+          )}
```
etojeDenys, line 50: what do you mean here, error can be something different than `'ERROR'`? looks like it can be boolean state and if `isError` === true then you can show the error message

src/components/People.tsx
```diff
@@ -0,0 +1,88 @@
...
+import { PersonLink } from './PersonLink';
+import { getPeopleWithParents } from '../utils/utils';
+
+export const People: React.FC = () => {
+  const [peopleFromServer, setPeopleFromServer] = useState<Person[] | null>(
+    null,
+  );
+  const [error, setError] = useState<string | null>(null);
+  const [isLoading, setIsLoading] = useState<boolean>(false);
+  const isArr = Array.isArray(peopleFromServer);
```
etojeDenys, line 14: looks redundant right now

etojeDenys (changes requested):

src/components/People.tsx
Solovei-Andrii: if I remove null from the types, then the message will be displayed for the first 0.5 seconds on boot, as the state will be checked, for example (`!error && peopleFromServer.length === 0 && { ... }`)
only when the state is renewed, `no people` will disappear

volodymyr-soltys97 (approved):
GJ!

PR #1582, opened 2024-08-13
add solution

om1gse:
[DEMO LINK](https://om1gse.github.io/react_people-table-basics/)


src/components/Navbar.tsx
```diff
@@ -0,0 +1,32 @@
...
+    className="navbar is-fixed-top has-shadow"
+    role="navigation"
+    aria-label="main navigation"
+  >
+    <div className="container">
+      <div className="navbar-brand">
+        <NavLink
+          className={({ isActive }) =>
+            `navbar-item ${isActive ? 'has-background-grey-lighter' : ''}`
+          }
```
alexander-ignatow, line 15: Noy critical - but consider using library like `classnames`

src/components/PersonRow.tsx
```diff
@@ -0,0 +1,56 @@
...
+export const PersonRow: React.FC<Props> = ({
+  person,
+  people,
+  handlePersonClick,
+  selectedPerson,
+}) => {
+  const { slug, sex, born, died, motherName, fatherName } = person;
+
+  const mother = people.find(p => p.name === motherName);
+  const father = people.find(p => p.name === fatherName);
```
alexander-ignatow, line 23: makes sense to calculate it once on a higher level, because now you are re-calculating these on each re-render

At least you can use `useMemo` on these values

alexander-ignatow (approved):
Good in overal, just mind some minor comments

PR #1604, opened 2024-09-01
Develop

NIKaragu:
[DEMO LINK](https://nikaragu.github.io/react_people-table-basics/)

### DIFF
--- a/src/components/PeopleTable/PersonRow/PersonRow.tsx
+++ b/src/components/PeopleTable/PersonRow/PersonRow.tsx
@@ -1,0 +1,55 @@
+import { useCallback, useMemo } from 'react';
+import { Person } from '../../../types';
+import { PersonLink } from './PersonLink';
+import classNames from 'classnames';
+
+type Props = {
+  personToRender: Person;
+  slug?: string;
+  people: Person[];
+};
+
+export const PeopleRow: React.FC<Props> = ({
+  personToRender,
+  slug,
+  people,
+}) => {
+  const findParentInTable = useCallback(
+    (parentName: string | null) => {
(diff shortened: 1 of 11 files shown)


src/pages/PeoplePage/PeoplePage.tsx
```diff
@@ -0,0 +1,56 @@
...
+  const emptyPeopleMessage = (
+    <p data-cy="noPeopleMessage">There are no people on the server</p>
+  );
+
+  useEffect(() => {
+    setArePeopleLoading(true);
+    getPeople()
+      .then(peopleFromServer => {
+        setPeople(peopleFromServer);
+        setArePeopleLoading(false);
```
iyehorova, line 26: You can put setArePeopleLoading(false); in `finaly` instead of using twice in `then` and `catch`

src/components/PeopleTable/PersonRow/PersonRow.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+type Props = {
+  personToRender: Person;
+  slug?: string;
+  people: Person[];
+};
+
+export const PeopleRow: React.FC<Props> = ({
+  personToRender,
+  slug,
```
iyehorova, line 14: You don't need to pass slug as a property, apply useSearchParams directly in the component

iyehorova (approved):
Hi! Good job!
Left some comments below

src/components/PeopleTable/PersonRow/PersonRow.tsx
NIKaragu: nice idea, but we haven't come to this topic yet)

PR #1623, opened 2024-09-06
add task solution

ogchy:
https://ogchy.github.io/react_people-table-basics/


src/types/Person.ts
```diff
@@ -7,5 +7,6 @@ export interface Person {
   motherName: string | null;
   slug: string;
   mother?: Person;
+  motherSlug?: string;
```
etojeDenys, line 10: looks redundant

etojeDenys (approved):
lgtm

PR #1665, opened 2024-10-24
Solution

Hradivnyk:
[DEMO LINK](https://Sholudyvyy.github.io/react_people-table-basics/)

### DIFF
--- a/src/components/PeopleTable/PeopleTable.tsx
+++ b/src/components/PeopleTable/PeopleTable.tsx
@@ -1,0 +1,47 @@
+import React from 'react';
+import { Person } from '../../types';
+import { PersonInfo } from '../PersonInfo/PersonInfo';
+
+type Props = {
+  people: Person[];
+};
+
+export const PeopleTable: React.FC<Props> = ({ people }) => {
+  return (
+    <table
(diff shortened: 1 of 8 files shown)


src/components/PersonInfo/PersonInfo.tsx
```diff
@@ -0,0 +1,50 @@
+import { Link, useParams } from 'react-router-dom';
+import { Person } from '../../types';
+import classNames from 'classnames';
+
+type Props = {
+  person: Person;
+  father?: Person;
+  mother?: Person;
+};
```
EustasTheMonk, line 9: We have the same type in Person type

src/components/PeopleTable/PeopleTable.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+          return (
+            <PersonInfo
+              key={person.slug}
+              person={person}
+              father={father}
+              mother={mother}
+            />
+          );
+        })}
```
EustasTheMonk, line 43: May be we should prepare people right after we get them since we have father and mother fields in Person type.

EustasTheMonk (approved):
Overall looks good!

PR #1667, opened 2024-10-24
Solution

vikshiyan:
[DEMO LINK](https://vikshiyan.github.io/react_people-table-basics/)

### DIFF
--- a/src/Root.tsx
+++ b/src/Root.tsx
@@ -1,0 +1,23 @@
+import { Navigate, Route, Routes } from 'react-router-dom';
+
+import { App } from './App';
+import { HomePage } from './components/HomePage';
+import { PageNotFound } from './components/PageNotFound';
+import { PeoplePage } from './components/PeoplePage';
+
+export const Root = () => {
+  return (
+    <Routes>
+      <Route path="/" element={<App />}>
+        <Route index element={<HomePage />} />
+        <Route path="/home" element={<Navigate to={'/'} />} />
+        <Route path="people">
+          <Route index element={<PeoplePage />} />
+          <Route path={`:slug?`} element={<PeoplePage />} />
+        </Route>
+
+        <Route path="*" element={<PageNotFound />} />
+      </Route>
+    </Routes>
+  );
+};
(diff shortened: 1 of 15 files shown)


src/Root.tsx
```diff
@@ -0,0 +1,23 @@
...
+import { HomePage } from './components/HomePage';
+import { PageNotFound } from './components/PageNotFound';
+import { PeoplePage } from './components/PeoplePage';
+
+export const Root = () => {
+  return (
+    <Routes>
+      <Route path="/" element={<App />}>
+        <Route index element={<HomePage />} />
+        <Route path="/home" element={<Navigate to={'/'} />} />
```
YuriiMykhailenko: Your solution doesn't follow requirement #3 from readme 

```suggestion
        <Route path="/home" element={<Navigate to={'/'} replace />} />
```

src/Root.tsx
```diff
@@ -0,0 +1,23 @@
...
+        <Route path="people">
+          <Route index element={<PeoplePage />} />
+          <Route path={`:slug?`} element={<PeoplePage />} />
```
YuriiMykhailenko: Why did you duplicate element here?

src/components/PeoplePage/PeoplePage.tsx
```diff
@@ -0,0 +1,118 @@
...
+          <div className="box table-container">
+            {loading && <Loader />}
+
+            {error && (
+              <p data-cy="peopleLoadingError" className="has-text-danger">
+                Something went wrong
+              </p>
+            )}
+
+            {people.length === 0 && !loading && !error && (
```
YuriiMykhailenko: It will be better to move it into a variable

```suggestion
            const example = !people.length && !loading && !error;
            
            {example && (
```

src/components/PeoplePage/PeoplePage.tsx
```diff
@@ -0,0 +1,118 @@
...
+                      if (!human) {
+                        return '-';
+                      }
+
+                      if (typeof human !== 'string') {
+                        return <PersonLink person={human} />;
+                      }
+
+                      return humanName;
+                    };
```
YuriiMykhailenko: You don't need that much code here, move functions outside of map and reconsider their logic you can make it times better. Instead of searching parents here make mapped version right after fetch(pay attention to person type there's a hint) and only then make setPeople.

YuriiMykhailenko (approved):
Almost there👍 Please consider my comments.

PR #1669, opened 2024-10-24
solved task

PikalovAndrey:
[DEMO LINK](https://PikalovAndrey.github.io/react_people-table-basics/)

### DIFF
--- a/src/PersonLink.tsx
+++ b/src/PersonLink.tsx
@@ -1,0 +1,22 @@
+import { FC } from 'react';
+import { Person } from './types';
+
+interface PersonLinkProps {
+  person: Person;
+  isMother?: boolean;
+}
+
+export const PersonLink: FC<PersonLinkProps> = ({ person, isMother }) => (
+  <>
+    {person.name ? (
+      <a
+        href={`#/people/${person.slug}`}
+        className={isMother ? 'has-text-danger' : ''}
+      >
+        {person.name}
+      </a>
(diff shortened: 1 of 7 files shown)


src/index.tsx
```diff
@@ -1,13 +1,28 @@
...
+import { HashRouter, Navigate, Route, Routes } from 'react-router-dom';
+import { HomePage } from './components/HomePage';
+import { PeoplePage } from './components/PeoplePage';
+import { NotFoundPage } from './components/NotFound';
 
 createRoot(document.getElementById('root') as HTMLDivElement).render(
-  <Router>
-    <App />
-  </Router>,
+  <html className="has-navbar-fixed-top">
```
Valeraom, line 13: you can add this class to the `html` tag in the index.html

src/PersonLink.tsx
```diff
@@ -0,0 +1,22 @@
...
+import { Person } from './types';
+
+interface PersonLinkProps {
+  person: Person;
+  isMother?: boolean;
+}
+
+export const PersonLink: FC<PersonLinkProps> = ({ person, isMother }) => (
+  <>
+    {person.name ? (
```
Valeraom, line 11: why do we need this condition?
if i understand correctly, this component can only get person as an object, because you've checked it in the `createPerson` function

Valeraom (approved):
Well done!

PR #1676, opened 2024-10-26
add solution

VladTk:
[DEMO LINK](https://VladTk.github.io/react_people-table-basics/)


src/utils/findPersonByName.ts
```diff
@@ -0,0 +1,12 @@
+import { Person } from '../types';
+
+export const findPersonByName = (
+  name: string | null,
+  people: Person[],
+): Person | undefined => {
+  if (!name) {
+    return undefined;
```
vadiimvooo, line 8: it's better to return null here

vadiimvooo (approved):
Good job, one small comment below.

PR #1735, opened 2024-12-26
solution

Fridrif27:
[DEMO LINK](https://fridrif27.github.io/react_people-table-basics/)

### DIFF
--- a/src/components/PersonItem.tsx
+++ b/src/components/PersonItem.tsx
@@ -1,0 +1,48 @@
+import React from 'react';
+import { Person } from '../types';
+import { Link, useParams } from 'react-router-dom';
+
+type Props = {
+  person: Person;
+};
+const PersonItem: React.FC<Props> = ({ person }) => {
+  const { slug } = useParams();
+
+  return (
+    <tr
+      data-cy="person"
+      className={person.slug === slug ? 'has-background-warning' : ''}
+    >
+      <td>
+        <Link
+          to={`${person.slug}`}
+          className={person.sex === 'f' ? 'has-text-danger' : ''}
+        >
+          {person.name}
+        </Link>
+      </td>
+
+      <td>{person.sex}</td>
+      <td>{person.born}</td>
+      <td>{person.died}</td>
+      <td>
+        {person.mother ? (
+          <Link to={person.mother.slug} className="has-text-danger">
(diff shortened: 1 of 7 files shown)


src/components/PersonItem.tsx
```diff
@@ -0,0 +1,48 @@
...
+type Props = {
+  person: Person;
+};
+const PersonItem: React.FC<Props> = ({ person }) => {
+  const { slug } = useParams();
+
+  return (
+    <tr
+      data-cy="person"
+      className={person.slug === slug ? 'has-background-warning' : ''}
```
taniabarkovskya, line 14: you can try use classNames here, because you don't have second option

src/components/PersonItem.tsx
```diff
@@ -0,0 +1,48 @@
...
+    >
+      <td>
+        <Link
+          to={`${person.slug}`}
+          className={person.sex === 'f' ? 'has-text-danger' : ''}
```
taniabarkovskya, line 19: you can try use classNames here, because you don't have second option

src/components/PersonItem.tsx
```diff
@@ -0,0 +1,48 @@
...
+          <Link to={person.mother.slug} className="has-text-danger">
+            {person.motherName}
+          </Link>
+        ) : (
+          person.motherName || '-'
+        )}
+      </td>
+      <td>
+        {person.father ? (
+          <Link to={person.father.slug}>{person.father.name}</Link>
```
taniabarkovskya, line 39: you use here father.name but in previous line - motherName. I may be wrong, but I think it's better to use similar notation for these outputs

src/pages/PeoplePage.tsx
```diff
@@ -0,0 +1,94 @@
...
+            const mam = result.find(p => p.name === person.motherName);
+            const dad = result.find(p => p.name === person.fatherName);
+
+            if (mam) {
+              modifiedPerson.mother = mam;
+            }
+
+            if (dad) {
+              modifiedPerson.father = dad;
+            }
```
taniabarkovskya, line 28: you can delete this because find will return undefined anyway

taniabarkovskya (approved):
good job, I`m approving this! 💪
but please pay attention to comments :)

PR #1754, opened 2024-12-30
Develop

SemenVodolazskij:
[DEMO LINK](https://SemenVodolazskij.github.io/react_people-table-basics/)


src/components/PeopleTable.tsx
```diff
@@ -0,0 +1,95 @@
...
+              </Link>
+            </td>
+
+            <td>{person.sex}</td>
+            <td>{person.born}</td>
+            <td>{person.died}</td>
+            <td>
+              {person.motherName ? (
+                <>
+                  {isPersoninList(person.mother?.slug) ? (
```
AnderZerfall, line 60: I suggest moving this and similar parts into a separate component, PersonLink, so that we can reuse it wherever we need without repeating the same code. As far as I know, the same thing was established in the task description

src/types/Person.ts
```diff
@@ -6,6 +6,9 @@ export interface Person {
   fatherName: string | null;
   motherName: string | null;
   slug: string;
-  mother?: Person;
-  father?: Person;
+}
```
AnderZerfall, line 9: Interesting solution, though not so necessary.  You could use a simple Person type too, and create some function (in utils folder for example) to define whether the person has any of their parents on the current list.

AnderZerfall (approved):
Good job!

PR #1767, opened 2025-01-05
solution

maxsabo:
[DEMO LINK](https://maxsabo.github.io/react_people-table-basics/)


Eeoneishin (changes requested):
You need to send advanced PR for review, and you sent basics

PR #1771, opened 2025-01-06
solution

so3r:
https://so3r.github.io/react_people-table-basics/


src/helpers/TableHeaders.ts
```diff
@@ -0,0 +1 @@
+export const TableHeaders = ['Name', 'Sex', 'Born', 'Died', 'Mother', 'Father'];
```
BudnikOleksii, line 1: Only classes, constructor functions and components should be in PascalCase

BudnikOleksii (approved):
LGTM 🚀

PR #1873, opened 2025-03-29
Implemented the App with 2 pages and ability to select a person in th…

oleksandr-kovalchuk:
[DEMO LINK](https://oleksandr-kovalchuk.github.io/react_people-table-basics/)

### DIFF
--- a/src/components/Navbar.tsx
+++ b/src/components/Navbar.tsx
@@ -1,8 +1,13 @@
+import classNames from 'classnames';
 import { NavLink } from 'react-router-dom';
 
+type NavLinkProps = {
+  isActive: boolean;
+};
+
 const Navbar = () => {
-  const getNavLinkClass = ({ isActive }: { isActive: boolean }) =>
-    `navbar-item ${isActive ? 'has-background-grey-lighter' : ''}`;
+  const getNavLinkClass = ({ isActive }: NavLinkProps) =>
+    classNames('navbar-item', { 'has-background-grey-lighter': isActive });
 
   return (
     <nav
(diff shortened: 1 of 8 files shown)


oleksandr-kovalchuk:
I am using HashRouter in the index.tsx file

src/components/PeopleTable.tsx
```diff
@@ -0,0 +1,72 @@
+import { Person } from '../types';
+import PersonLink from './PersonLink';
+
+type Props = {
+  people: Person[];
+  selectedSlug?: string;
+};
+
+const PeopleTable: React.FC<Props> = ({ people, selectedSlug }) => {
```
TarasHoliuk: nit: you can get slug from params here instead of passing it as a prop

src/components/Navbar.tsx
```diff
@@ -0,0 +1,28 @@
+import { NavLink } from 'react-router-dom';
+
+const Navbar = () => {
+  const getNavLinkClass = ({ isActive }: { isActive: boolean }) =>
```
TarasHoliuk: you might define it outside of the component

src/components/PersonLink.tsx
```diff
@@ -0,0 +1,25 @@
...
+  if (!person) {
+    return <span>-</span>;
+  }
+
+  const isFemale = person.sex === 'f';
+
+  return (
+    <NavLink
+      to={`/people/${person.slug}`}
+      className={`has-text-weight-bold ${isFemale ? 'has-text-danger' : ''}`}
```
TarasHoliuk: It's better to use classnames library (or smth similar) for such conditional classes

TarasHoliuk (approved):
Approved with a few small comments FYI ✔️

PR #1893, opened 2025-04-07
Develop

R-Ohirok:
[DEMO LINK](https://R-Ohirok.github.io/react_people-table-basics/)


src/components/PeoplePage/PeoplePage.tsx
```diff
@@ -0,0 +1,59 @@
...
+      })
+      .catch(() => {
+        setIsError(true);
+      })
+      .finally(() => {
+        setIsLoading(false);
+      });
+  }, []);
+
+  if (isLoading) {
```
PiekharievaK, line 37: You can move the loader to the container

```
<main className="section">
      <div className="container">
      <Loader />
      ...
```
      
Because now it appears on the page, but it is not visible, it is covered by navigation

PiekharievaK (approved):
good job

PR #1907, opened 2025-04-08
add solution

vladrlex:
[DEMO LINK](https://vladrlex.github.io/react_people-table-basics/)


src/pages/PeoplePage.tsx
```diff
@@ -0,0 +1,55 @@
+import React, { useEffect, useState } from 'react';
+import { getPeople } from '../api';
+import { Person } from '../types';
+import { Loader } from '../components/Loader';
+import { PeopleTab } from '../components/PeopleTab';
+import { useParams } from 'react-router-dom';
+
+export const PeoplePage = () => {
+  const [isLoading, setIsLoading] = useState(false);
+  const [error, setError] = useState<string | null>(null);
```
daniil-trusov, line 10: You only use `error` as boolean value, so string type is not needed here, you can change it to `const [error, setError] = useState(false)`

src/components/PeopleTab.tsx
```diff
@@ -0,0 +1,56 @@
...
+        {people.map(person => (
+          <tr
+            key={person.slug}
+            data-cy="person"
+            className={
+              person.slug === selectedSlug ? 'has-background-warning' : ''
+            }
+          >
+            <td>
+              <PersonLink name={person.name} people={people} />
```
daniil-trusov, line 40: You already have a person with all data, why are you sending only the name to `PersonLink` and then searching for the person again? Instead, you can send the person to `PersonLink`. For mother and father - you can find them here in `.map()` and send their details to `PersonLink`:

```
father = people.find(f => f.name = person.fatherName);
<PersonLink person={father} />
```

daniil-trusov (approved):
Good job, however you can improve the efficiency of the solution - check the comments to see how it can be done.

PR #1912, opened 2025-04-10
Dev

Iwantschenko:
[DEMO LINK](https://Iwantschenko.github.io/react_people-table-basics/)


VitaliiHutsalo (approved):
Well Done!!!!

src/Root.tsx
```diff
@@ -0,0 +1,27 @@
...
+      <Routes>
+        <Route path={RoutesParts.root} element={<App />}>
+          <Route index element={<HomePage />} />
+          <Route
+            path={RoutesParts.home}
+            element={<Navigate to={RoutesParts.root} replace />}
+          />
+          <Route path={RoutesParts.people}>
+            <Route index element={<PeoplePage />} />
+            <Route path={RoutesParts.personId} element={<PeoplePage />} />
```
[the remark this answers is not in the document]
alexander-ignatow: > The current configuration renders the entire `PeoplePage` component for both the `/people` route and the `/people/:slug` route. To correctly display individual person details, you should create a separate component (e.g., `PersonDetailsPage`) to render when a `personId` is provided in the URL. Then, render `PersonDetailsPage` here instead of `PeoplePage`.

Ignore this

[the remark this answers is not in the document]
alexander-ignatow: > The route with `personId` should render a separate component to display person details, not the whole `PeoplePage`.

Ignore this

src/types/RoutesURLParts.ts
```diff
@@ -0,0 +1,7 @@
+export const RoutesParts = {
+  root: '/',
+  home: '/home',
+  people: '/people',
+  personId: ':slug',
+  notFount: '*',
+};
```
alexander-ignatow, line 7: Few issues here

1. This is not a type - this is a constant
2. Even if used like a constant - it is redundant. it does not make your code more readable.
I would suggest removing this file and simply inlining the values in the router

src/pages/PeoplePage/PeoplePage.tsx
```diff
@@ -0,0 +1,61 @@
...
+    if (!peopleList) {
+      return <Loader />;
+    }
+
+    if (!peopleList.length) {
+      return <p data-cy="noPeopleMessage">There are no people on the server</p>;
+    }
+
+    return <PeopleTable peopleList={peopleList} />;
+  };
```
alexander-ignatow: This is an anti-pattern - you effectively make a component inside the component

Remove this function an inline it's content

src/pages/PeoplePage/PeoplePage.tsx
```diff
@@ -0,0 +1,61 @@
...
+  useEffect(() => {
+    callRequest();
+  }, []);
+
+  return (
+    <div className="container">
+      <h1 className="title">People Page</h1>
+
+      <div className="block">{renderContent()}</div>
+    </div>
```
alexander-ignatow: maybe, some of these components may go upper component three - to the `App` ?

src/pages/HomePage/HomePage.tsx
```diff
@@ -0,0 +1,7 @@
+export const HomePage = () => {
+  return (
+    <div className="container">
+      <h1 className="title">Home Page</h1>
+    </div>
```
alexander-ignatow: The same code as in the PeoplePage - it makes sense to move it up the tree - to the `App`

alexander-ignatow (changes requested):
Good 👍 , but some refactoring is needed.
see the issues bellow.

src/types/RoutesURLParts.ts
Iwantschenko: I don't really understand what the problem is here.

alexander-ignatow: 1. RoutesParts is the set of the constants, not the type. So at the very least, it should be moved away from the types
2. Ok, you can keep it, if you want, but then rename it to something more meaningful (ex: RouteNames, PathNames, etc) and move away from the types folder (say to misc, utils, or constants)

src/pages/PeoplePage/PeoplePage.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+      setPeopleList(deepCopy);
+    } catch {
+      setErrorMessage(true);
+    }
+  };
+
+  useEffect(() => {
+    callRequest();
+  }, []);
```
alexander-ignatow, line 32: Why did you move callRequest to a separate function?

```suggestion
  useEffect(() => {
  const callRequest = async () => {
    try {
      const peopleFromServer = await getPeople();
      const deepCopy = peopleFromServer.map(person => ({
        ...person,
        mother: peopleFromServer.find(
          mother => mother.name === person.motherName,
        ),
        father: peopleFromServer.find(
          father => father.name === person.fatherName,
        ),
      }));

      setPeopleList(deepCopy);
    } catch {
      setErrorMessage(true);
    }
  };

    callRequest();
  }, []);
```

Why do so?
- Now linter can check your code for the correct useEffect dependencies
- No need to re-cereate the function on each component render
- If the function is relatively simple (like yours) the code remains readable

alexander-ignatow (approved):
Looks better now!

Let's move on. Still, see my comments to avoid the issues in the future

PR #1918, opened 2025-04-13
Develop

nex1994:
[DEMO LINK](https://nex1994.github.io/react_people-table-basics/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,46 +1,9 @@
-import { Loader } from './components/Loader';
 import './App.scss';
-import {
-  Navigate,
-  NavLink,
-  Outlet,
-  Route,
-  Routes,
-  useLocation,
-} from 'react-router-dom';
-import { getPeople } from './api';
+import { Navigate, NavLink, Outlet, Route, Routes } from 'react-router-dom';
 import { PeopleTable } from './components/PeopleTable';
-import { useEffect, useState } from 'react';
-import { Person } from './types';
 import classNames from 'classnames';
 
-const STATUS = {
-  resolved: 'resolved',
-  rejected: 'rejected',
-  idle: 'idle',
-  pending: 'pending',
-} as const;
-
-type Status = (typeof STATUS)[keyof typeof STATUS];
-
 export const App = () => {
-  const [people, setPeople] = useState<Person[] | null>(null);
-  const [status, setStatus] = useState<Status>('idle');
-  const { pathname } = useLocation();
-  const loadPeople = () => {
-    return getPeople()
-      .then(data => {
-        setPeople(data);
-        setStatus('resolved');
-      })
-      .catch(() => setStatus('rejected'));
-  };
-
-  useEffect(() => {
-    setStatus('pending');
-    loadPeople();
-  }, []);
-
   return (
     <>
       <div data-cy="app">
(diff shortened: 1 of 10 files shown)


src/App.tsx
```diff
@@ -1,167 +1,120 @@
...
+        setPeople(data);
+        setStatus('resolved');
+      })
+      .catch(() => setStatus('rejected'));
+  };
+
+  useEffect(() => {
+    setStatus('pending');
+    loadPeople();
+  }, []);
```
zibi-dev: API request for data should be done in PeopleTable component

src/App.tsx
```diff
@@ -1,167 +1,120 @@
...
+                  </p>
+                )}
+
+                {people?.length === 0 && (
+                  <p data-cy="noPeopleMessage">
+                    There are no people on the server
+                  </p>
+                )}
+              </div>
+            </div>
```
zibi-dev: This logic should be within the PeopleTable component. Then you won't have to check if the pathname is correct. React Router handles that for you within the Routes component.

src/App.tsx
```diff
@@ -1,167 +1,120 @@
...
+              >
+                People
+              </NavLink>
+            </div>
+          </div>
+        </nav>
+
+        <main className="section">
+          <div className="container">
+            {pathname === '/home' && <Navigate to="/" replace={true} />}
```
zibi-dev: You can do it with Route
```suggestion
            <Route path="/home" element={<Navigate to="/" replace={true} />} />
```

src/App.tsx
```diff
@@ -1,167 +1,120 @@
...
+            <Routes>
+              <Route path="/" element={<h1 className="title">Home Page</h1>} />
+              {people?.length !== 0 && status === 'resolved' && (
+                <Route path="/people" element={<PeopleTable people={people} />}>
+                  <Route path=":personSlug" element={<Outlet />} />
+                </Route>
+              )}
```
zibi-dev: Yes, that's generally considered bad practice. Routes should be statically defined whenever possible. Dynamically adding/removing routes based on data can lead to unexpected behavior, especially with features like route matching.

src/components/PeopleTable/PeopleTable.tsx
```diff
@@ -0,0 +1,42 @@
+import { Person } from '../../types';
+import { PersonRow } from '../PersonRow';
+
+type Props = {
+  people?: Person[] | null;
+};
+
+export const PeopleTable = ({ people }: Props) => {
+  return (
```
zibi-dev, line 34: Here should be stored all logic for fetching data or even you can create other component for example PeoplePage

src/components/PersonRow/PersonRow.tsx
```diff
@@ -0,0 +1,55 @@
...
+  const activeRowClassName = classNames({
+    'has-background-warning': pathname === `/people/${person?.slug}`,
+  });
+  const personMother: Person | undefined = people.find(
+    searchedPerson => searchedPerson.name === person?.motherName,
+  );
+
+  const personFather: Person | undefined = people.find(
+    searchedPerson => searchedPerson.name === person?.fatherName,
+  );
```
zibi-dev: You don't have explicitly type everything. This will be properly infered.

Zibi95 (changes requested):
Hey. Good job. There some things to fix.

Zibi95 (approved):

PR #1921, opened 2025-04-14
Solution

Olehgoon093:


alexander-ignatow (changes requested):
Good job, but improvements are needed

1. Please add the deployment link
2. Clean up all warnings (unused imports and so on)
3. Remove commented code

overall - it looks like the solution submitted is incomplete

PR #1982, opened 2025-06-21
add task solution

AndSLASH:
[DEMO LINK](https://andslash.github.io/react_people-table-basics/)


src/components/People/People.tsx
```diff
@@ -0,0 +1,42 @@
...
+          <th>Born</th>
+          <th>Died</th>
+          <th>Mother</th>
+          <th>Father</th>
+        </tr>
+      </thead>
+
+      <tbody>
+        {people.map(person => {
+          return (
```
nbryk: It's not mistake, but in this case { return } is no longer needed. 😉

src/components/Person/Person.tsx
```diff
@@ -0,0 +1,88 @@
...
+        className={cn({
+          'has-background-warning': isSelected,
+        })}
+      >
+        <td>
+          <PersonLink
+            to={`/people/${person.slug}`}
+            displayText={person.name}
+            isFemale={person.sex === 'f'}
+          />
```
nbryk: * create the `<PersonLink person={person} />` and use it for all existing people;
Perhaps we could just pass 'person' to the component and handle the gender check inside it?

nbryk (approved):
Overall, the work is excellent. Well done! 👍

PR #2096, opened 2025-09-05
add solution

YuliaProkipchuk:
[DEMO LINK](https://YuliaProkipchuk.github.io/react_people-table-basics/)


AARON-FOX (approved):
Great solution🔥

YuliaProkipchuk:
@mateacademy-ai-mentor a <Link to={parent.slug}> within a nested route /people works correctly because it's interpreted as a relative path, resolving to /people/${parent.slug} without requiring the full, absolute path to be explicitly written. Also there is no need to do selection with state, because i do it with the URL parameter and after reloading page highlighted row must remain in the same state, so in this case using useState won't be correct. And there was no specification in requirements about how exactly to implement this logic: **Highlight the row of the selected person with the has-background-warning class;**

YuliaProkipchuk:
@mateacademy-ai-mentor I have the RELATIVE path, i don't need to use there the absolute path: `<Route path="people" element={<PeoplePage />}>
          <Route path=":personId?" element={<PeoplePage />} />
        </Route>`

PR #2100, opened 2025-09-05
add task solution

jeolinna:
[DEMO LINK](https://jeolinna.github.io/react_people-table-basics/)


src/components/PeopleTable/PeopleTable.tsx
```diff
@@ -0,0 +1,66 @@
...
+        </tr>
+      </thead>
+      <tbody>
+        {people.map(person => {
+          const mother = person.motherName
+            ? people.find(p => p.name === person.motherName) || null
+            : null;
+          const father = person.fatherName
+            ? people.find(p => p.name === person.fatherName) || null
+            : null;
```
derpedcatto, line 34: `Person` interface has fields for `mother?` and `father?` objects, and it would be better to prepare whole `people` array without having to do logic in lower components.
```ts
const preparePeople = (people: Person[]): Person[] => {
  return people.map(person => ({
    ...person,
    mother: people.find(mother => mother.name === person.motherName),
    father: people.find(father => father.name === person.fatherName),
  }));
};
```

src/components/PersonLink/PersonLink.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+type Props = {
+  person: Person | null;
+  name?: string | null;
+};
+
+export const PersonLink: React.FC<Props> = ({ person, name }) => {
+  if (!name) {
+    return <>-</>;
+  }
```
derpedcatto, line 13: Small nitpick, but it could cause confusion if component `PersonLink` would not return an actual link.

I think it's better to have something like this in table row from my example (even though it's more code, at least someone else looking at this code will instantly know what exactly will be placed in <td> and why):
```tsx
            <td>
              {person.mother ? (
                <PersonLink person={person.mother} />
              ) : (
                person.motherName || '-'
              )}
            </td>
            <td>
              {person.father ? (
                <PersonLink person={person.father} />
              ) : (
                person.fatherName || '-'
              )}
            </td>
            ```

File: src/pages/PeoplePage.tsx [hunk 3]
```diff
@@ -0,0 +1,48 @@
...
+
+          {!isLoading && !error && people.length > 0 && (
+            <PeopleTable people={people} selectedSlug={slug} />
+          )}
+
+          {!isLoading && !error && people.length === 0 && (
+            <p data-cy="noPeopleMessage">There are no people on the server</p>
+          )}
+        </div>
+      </div>
```
derpedcatto, line 43: Here is an alternative way you could do conditional rendering without re-use of these checks:
```tsx
export const PeoplePage = () => {

// ...

  if (isLoading) {
    return <Loader />;
  }

  if (isError) {
    return (
      <p data-cy="peopleLoadingError" className="has-text-danger">
        Something went wrong
      </p>
    );
  }

  return (
    <>
      <h1 className="title">People Page</h1>

      <div className="block">
        <div className="box table-container">
          {people.length === 0 ? (
            <p data-cy="noPeopleMessage">There are no people on the server</p>
          ) : (
            <PeopleTable people={people} selectedPersonSlug={personSlug} />
          )}
        </div>
      </div>
    </>
  );
};
```

derpedcatto (approved):
Good job!

PR #2101, opened 2025-09-05
add solution

derpedcatto:
[DEMO LINK](https://derpedcatto.github.io/react_people-table-basics/)


Oleksandr-Oliferuk (approved):
Great Job!

derpedcatto:
Yes, API provides only names, but remember that provided `Person` interface looks like this:
```ts
export interface Person {
  name: string;
  sex: string;
  born: number;
  died: number;
  fatherName: string | null;
  motherName: string | null;
  slug: string;
  mother?: Person;
  father?: Person;
}
```
And I prepare API data in Page component, so `mother?` and `father?` gets assigned correct values and it gets passed down below. My app is working as intended and I already implemented your first solution, although I do it in `Page` instead of `Table` component.

alexander-ignatow (approved):
👍

PR #2117, opened 2025-09-10
solution

DidZhara:
[DEMO LINK](https://DidZhara.github.io/react_people-table-basics/)


bebelekele (approved):
Good job. You also can add a `class="has-navbar-fixed-top"` to `<html>` in index.html to make it looks better.

PR #2180, opened 2025-11-07
add solution

dashasssss:
https://dashasssss.github.io/react_people-table-basics/


MarinaKramarchuk (approved):
good job! 
if you want the pages to have indents like in the example, add the following class to the index.html file:
html lang="en" class="has-navbar-fixed-top"

PR #2183, opened 2025-11-07
feat: implement people table functionality

MariaAlefirenko:
[DEMO LINK](https://MariaAlefirenko.github.io/react_people-table-basics/)

### DIFF
--- a/src/components/Pages/PeoplePage.tsx
+++ b/src/components/Pages/PeoplePage.tsx
@@ -49,99 +49,87 @@ export const PeoplePage: React.FC = () => {
   const hasPeople = !isLoading && !error && people.length > 0;
 
   return (
-      <div className="container">
-        <h1 className="title">People Page</h1>
+    <div className="container">
+      <h1 className="title">People Page</h1>
 
-        <div className="block">
-          <div className="box table-container">
-            {isLoading && (
-              <div data-cy="loader">
-                <Loader />
-              </div>
-            )}
+      <div className="block">
+        <div className="box table-container">
+          {isLoading && (
+            <div data-cy="loader">
+              <Loader />
+            </div>
+          )}
 
-            {error && (
-              <p data-cy="peopleLoadingError" className="has-text-danger">
-                {error}
-              </p>
-            )}
+          {error && (
+            <p data-cy="peopleLoadingError" className="has-text-danger">
+              {error}
+            </p>
+          )}
 
-            {noPeople && (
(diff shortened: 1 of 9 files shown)


src/components/PersonLink.tsx
```diff
@@ -0,0 +1,49 @@
...
+      personName?: never;
+      people?: never;
+    }
+  | {
+      personName: string | null;
+      people: Person[];
+      person?: never;
+    };
+
+export const PersonLink: React.FC<Props> = props => {
```
voloshynmykyta: Read something about [KISS](https://en.wikipedia.org/wiki/KISS_principle)

You can make just something like this: 

```js
import React from 'react';
import { Person } from '../../types';
import { Link } from 'react-router-dom';
import cn from 'classnames';

interface PersonLinkProps {
  person?: Person | null;
  name?: string | null;
}

export const PersonLink: React.FC<PersonLinkProps> = ({ person, name }) => {
  if (person) {
    return (
      <Link
        to={`/people/${person.slug}`}
        className={cn({ 'has-text-danger': person.sex === 'f' })}
      >
        {name || person.name}
      </Link>
    );
  }

  if (name) {
    return <>{name}</>;
  }

  return <>-</>;
};
```
And use it in more simple way in people table: 

`<PersonLink person={person.mother} name={person.motherName} />`


it will remove a lot of different condtition renders in your people table.

src/components/Pages/PeoplePage.tsx
```diff
@@ -0,0 +1,147 @@
...
+
+                    const byName = new Map(people.map(p => [p.name, p]));
+                    const mother = motherName
+                      ? byName.get(motherName)
+                      : undefined;
+                    const father = fatherName
+                      ? byName.get(fatherName)
+                      : undefined;
+
+                    return (
```
voloshynmykyta: I would like to see here another one separated component with more clear rendering(less conditions)

src/components/PersonLink.tsx
```diff
@@ -0,0 +1,49 @@
+/* eslint-disable @typescript-eslint/indent */
+import { Link, useParams } from 'react-router-dom';
+import { Person } from '../types/Person';
+
+type Props =
```
voloshynmykyta: in general, i would recommend another approach

but at least make two different types and use here instead of { ... } | { ... }

voloshynmykyta (changes requested):

PR #2194, opened 2025-11-10
add task solution

Julia-Hertseva:
[DEMO LINK](https://Julia-Hertseva.github.io/react_people-table-basics/)


andriiLink (approved):
Great job! To make your page look the same as on example just find in project index.html (it's outside src folder) and add class `class="has-navbar-fixed-top` to `html` tag

PR #2207, opened 2025-11-14
develop

ElinaMrachkovska:
https://elinamrachkovska.github.io/react_tabs-with-router/#/tabs/tab-2


ElinaMrachkovska:
explain where the errors are in the code, why the task is not accepted

PR #2279, opened 2026-02-06
add task solution

AdelinaY18:
[DEMO LINK](https://AdelinaY18.github.io/react_people-table-basics/)


Etwaiz (approved):
everything is fine, but I see that the tests are not passing, change them

PR #2337, opened 2026-03-19
Develop

AndreyKagaml:
[DEMO LINK](https://AndreyKagaml.github.io/react_people-table-basics/)

My app works, but tests pass and fail very strange


brespect (changes requested):
Good progress, but you need to pass all tests:
the error simply shows that your UI logic is not working properly.

**should not show loading error if API send no people**

That is:

- The API returned a successful response
- but the list of people is empty
- and you still show:

`<p class="has-text-danger">...</p>`

Cypress says: error should not exist, but it does

PR #2344, opened 2026-03-28
Added solution

MykolaFatkullin:


MykolaFatkullin:
@mateacademy-ai-mentor Link with slug is nested for people. So in this case "../" directs us to people

PR #2366, opened 2026-04-27
Implement the App with 2 pages and ability to select a person in the table

Dianaeva:
[DEMO LINK](https://dianaeva.github.io/react_people-table-basics/)


src/components/PeopleTable/PersonTableItem.tsx
```diff
@@ -0,0 +1,57 @@
...
+const PersonTableItemBase: React.FC<Props> = ({ person }) => {
+  const { sex, born, died, motherName, fatherName, slug, mother, father } =
+    person;
+  const { slug: paramSlug } = useParams();
+
+  const motherInfoTd = useMemo(() => {
+    if (!!motherName && !!mother) {
+      return <PersonLink person={mother} />;
+    } else if (!!motherName) {
+      return <>{motherName}</>;
```
polosanya, line 27:
```suggestion
    if (!!motherName && !!mother) {
      return <PersonLink person={mother} />;
    }
    
    if (!!motherName) {
      return <>{motherName}</>;
```
The code will not go further after `return` so there is no need to use `else` for better readability

src/components/PeopleTable/PersonTableItem.tsx
```diff
@@ -0,0 +1,57 @@
...
+      return <>-</>;
+    }
+  }, [motherName, mother]);
+
+  const fatherInfoTd = useMemo(() => {
+    if (!!fatherName && !!father) {
+      return <PersonLink person={father} />;
+    } else if (!!fatherName) {
+      return <>{fatherName}</>;
+    } else {
```
polosanya: Same here.
Also, as for me, it would be cleaner to check from the opposite:
`if (!fatherName && !father) {
      return <>-</>;
    }
`

src/pages/PeoplePage.tsx
```diff
@@ -0,0 +1,68 @@
...
+      .finally(() => {
+        setIsLoading(false);
+      });
+  }, []);
+
+  const extendedPeopleData = useMemo(() => {
+    if (people.length === 0) {
+      return people;
+    } else {
+      return [...people].map(pers => {
```
polosanya:
```suggestion
    if (people.length === 0) {
      return people;
    }
    
    return people.map(pers => {
```
`map` is a non-mutating method, no need to create a copy of array

src/pages/PeoplePage.tsx
```diff
@@ -0,0 +1,68 @@
...
+        <div className="box table-container">
+          {isLoading && <Loader />}
+
+          {errorMessage && (
+            <p data-cy="peopleLoadingError" className="has-text-danger">
+              {errorMessage}
+            </p>
+          )}
+
+          {!isLoading && !errorMessage && people.length === 0 && (
```
polosanya: Would be great to create a variable for this condition in case you need to update it later and for better readability

polosanya (changes requested):
Almost perfect! Let's improve you code a bit

polosanya (approved):

PR #2389, opened 2026-05-23
add solution

George-Buzhenitsa:
https://George-Buzhenitsa.github.io/react_people-table-basics/


Denys-Kravchuk9988 (approved):
Looks good to me!

Not critical is that there should be gap between header and content as on example. But it wasn't added in requirements.

Image

Image

PR #2408, opened 2026-06-01
add task solution

rlysunets:
[DEMO LINK](https://rlysunets.github.io/react_people-table-basics/)


src/App.tsx
```diff
@@ -1,167 +1,59 @@
...
-    </main>
-  </div>
-);
+      </nav>
+
+      <main className="section">
+        <div className="container">
+          <Routes>
+            <Route path="/" element={<HomePage />} />
+            <Route path="/people" element={<PeoplePage />} />
```
Palfinger2002, line 50: You can wrap both Route with similar way - "people" in one Route with way "people" and first with index and the other just ":slug". But it's up for you

Palfinger2002 (approved):
Good job!

